PR #83, opened 2015-02-16
Return prettyprinted JSON strings for errors.

hobeone:
This makes gomega pretty print both the actual and expected strings on an error.  I find that it's easier to see the differences this way.


onsi:
ha! durp - that was the whole point of `abuf` and `ebuf`!

onsi:
thanks :)

PR #89, opened 2015-03-30
Improve Succeed() failure message.

alext:
This adds the `Error()` output to the failure message to give better
information about the error that did occur.  This also makes the output
more similar to the negative failure message from the `HaveOccurred()`
matcher.


onsi:
looks good to me - thanks!

PR #91, opened 2015-04-04
file based matchers

james-lawrence:
Adds the following Matchers:
- BeAnExistingFile (succeeds for any type of file if it exists, regular, sym, directory, socket, etc)
- BeARegularFile (succeeds only for regular files)

Other possible matchers that could be created:
- BeADirectory
- BeASymlink (compatability probably is an issue, i.e. unix/darwin only)
- BeASocket (compatability probably is an issue, i.e. unix/darwin only)


onsi:
this looks great @james-lawrence -- thanks :)

I think `BeADirectory` would be a valuable addition too and would complement `BeADirectory`.

Any chance you'd be interested in updating the markdown docs on the gh-pages branch?

james-lawrence:
Absolutely tomorrow morning I'll see about doing both.

PR #94, opened 2015-04-16
Make usage of Succeed matcher slightly clearer

goonzoid:
A couple of times now I've heard people asking about exactly how this works, and each time I've taken a multi step trip through the docs, ending up in the `Handling Errors` section where I learn that matchers are always passed the first parameter to `Expect` or `Ω`.

I think this change would make that lookup a single step process, and is a little bit clearer.


onsi:
LGTM!

PR #96, opened 2015-04-30
Can use nil pointers that conform to error interface

craigfurman:
Hi Onsi,

We have found that it was not possible to use custom error types in negative assertions about errors. The problem can be observed using the example code:

```
var err *SomeCustomErrorType = nil
Ω(err).ShouldNot(HaveOccurred()) // panics
```

Thanks,
Craig Furman & George Lestaris


onsi:
Thanks @craigfurman

PR #98, opened 2015-05-19
Set Content-Type to application/json in RespondWithJSON* methods

matt-royal:
RespondWithJSONEncoded and RespondWithJSONEncodedPtr now set the response Content-Type header correctly.


onsi:
highlarious

PR #102, opened 2015-07-28
Show diff when MatchJSONMatcher fails

kwapik:


kwapik:
Don't have time to fix it now

PR #107, opened 2015-09-05
Added Implements() matcher to verify value implements a given interface

jim-slattery-rs:
- This is different from `BeAssignableToTypeOf()` because it matches an interface rather than a concrete type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .DS_Store
 *.test
 .
+# editor
+.idea
+*.iml
--- a/matchers.go
+++ b/matchers.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/onsi/gomega/matchers"
 	"github.com/onsi/gomega/types"
+	"reflect"
 )
 
 //Equal uses reflect.DeepEqual to compare actual with expected.  Equal is strict about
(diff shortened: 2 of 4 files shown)


jim-slattery-rs:
looks like the last commit hit an intermittent failure in the build, i don't have access to rerun it:
```
[Fail] Session signal [It] should send the signal to the command 
/home/travis/gopath/src/github.com/onsi/gomega/gexec/session_test.go:126
```

onsi:
I'm a little less convinced of this one, @jim-slattery-rs 

I tend to prefer to write tests like:

```
It("satisfies the Foo interface", func() {
})

**Comment by jim-slattery-rs** [comment]
I originally needed it to act as a guard when using `WithTransform()`, if the transform function expected input of a certain type, like:
```go
And(
  Implements(errorInterface),
  WithTransform(func(e interface{}) interface{} { return e.(error).Error() }, ...))
```
But since you suggested changing `WithTransform()` to use transform functions with explicit types, there may not be the same need for `Implements()`.. I will finish up those `WithTransform()` changes and see!

onsi:
Ah makes sense.  Yes, I think you'll find that the go runtime will throw a meaningful error if you start passing the wrong types around.  Thanks for your work on all this!

jim-slattery-rs:
Yep, `WithTransform()` is looking good in https://github.com/onsi/gomega/pull/108.
I'm glad to have learned how to pass around interface types in Go, even if the syntax is weird. But until I come across another use case for `Implements()`, I'll close this PR.

Issue #66, opened 2014-10-13
Return diff(s) of 2 strings


truongsinh:
Using one of diff libraries, gomega should return a diff of 2 strings, especially handy for comparing 2 lengthy strings.
- https://github.com/sergi/go-diff
- https://github.com/aryann/difflib

robdimsdale:
@truongsinh if this is still something you're interested in, you may be interested to know there is a PR open for this at https://github.com/onsi/gomega/pull/109

muhqu:
To whom it may concern, I created my own matcher to create colored unified and/or word diffs using the `git diff` tool. e.g. `worddiff.MatchJSON(..)`

See https://github.com/muhqu/gomega-additions

onsi:
Closing this out in favor of the PR

PR #109, opened 2015-09-16
Append diff of actual and expected to failure messages.

raphael:
When both actual and expected are strings and the diff tool is available
on the system.

This is not the prettyiest implementation (relies on diff and creates temp files) but
as it turns out implementing an efficient diff algorithm is not that simple ;)
However having a diff printed is the number one feature I am missing from rspec,
it makes finding problems a lot easier in cases the strings are long (e.g. JSON
or XML content).

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -5,6 +5,9 @@ package format
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
 	"reflect"
 	"strconv"
 	"strings"
@@ -45,7 +48,15 @@ var timeType = reflect.TypeOf(time.Time{})
 //The default indentation string emitted by the format package
 var Indent = "    "
 
+// SupportsDiff is true if there is a "diff" tool in the system PATH, false otherwise.
+var SupportsDiff bool
+
 var longFormThreshold = 20
+
+func init() {
+	_, err := exec.LookPath("diff")
+	SupportsDiff = (err == nil)
+}
 
 /*
 Generates a formatted matcher success/failure message of the form:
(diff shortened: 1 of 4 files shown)


robdimsdale:
@raphael out of curiosity could you post an example of what the new output might be some for some example xml or json? Does it depend on the implementation of the `diff` tool?

robdimsdale:
Also this looks like it might address https://github.com/onsi/gomega/issues/66

truongsinh:
+1

raphael:
Here is an example of a test that failed for a tool that generates Go code:
```
• Failure [0.003 seconds]
code generation
/home/raphael/go/src/github.com/raphael/goa/codegen/types_test.go:411
  Marshaler
  /home/raphael/go/src/github.com/raphael/goa/codegen/types_test.go:410
    with a simple object
    /home/raphael/go/src/github.com/raphael/goa/codegen/types_test.go:261
      generates the marshaler code [It]
      /home/raphael/go/src/github.com/raphael/goa/codegen/types_test.go:256

      Expected
          <string>: 
                if err == nil {
                        tmp1 := map[string]interface{}{
                                "foo": raw.Foo,
                        }
                        p = tmp1
          }
      to equal
          <string>:     p = map[string]interface{}{
                        "foo": raw.Foo,
                }
          
      
      diff:
      1,7c1,3
      < 
      <         if err == nil {
      <                 tmp1 := map[string]interface{}{
      <                         "foo": raw.Foo,
      <                 }
      <                 p = tmp1
      < }
      \ No newline at end of file
      ---
      >         p = map[string]interface{}{
      >                 "foo": raw.Foo,
      >         }
```
fwiw I found this extremely useful in my use cases, you can try it out by installing my fork.
The output does depend on the `diff` tool but that's a failrly stable dependency...

truongsinh:
+1

XenoPhex:
+1

robdimsdale:
@onsi I'd like to see this merged as well. I found myself wishing this existed the other day (in fact I re-wrote the test I was running to directly call diff).

I've merged the change locally (it's just a reordered import) and there is one test failure. @raphael can you run the tests locally and ensure they pass for you? I'm running OSX 10.11.2 with diff installed via brew:

```
[Fail] AndMatcher failure messages when match fails [It] gives a descriptive message
/Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:28

• Failure [0.017 seconds]
AndMatcher
/Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:103
  failure messages
  /Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:64
    when match fails
    /Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:56
      gives a descriptive message [It]
      /Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:55

      Expected
          <string>: Expected
              <string>: hi
          to equal
              <string>: hip

          diff:
          1c1
          < hi
          \ No newline at end of file
          ---
          > hip
          \ No newline at end of file

      to equal
          <string>: Expected
              <string>: hi
          to equal
              <string>: hip

      diff:
      4,12c4
      <     <string>: hip
      <
      < diff:
      < 1c1
      < < hi
      < \ No newline at end of file
      < ---
      < > hip
      < \ No newline at end of file
      ---
      >     <string>: hip
      \ No newline at end of file


      /Users/robdimsdale/go/src/github.com/onsi/gomega/matchers/and_test.go:28
------------------------------
```

```
diff --version
diff (GNU diffutils) 2.8.1
Copyright (C) 2002 Free Software Foundation, Inc.
```

scottschulthess:
This would be huge

raphael:
@robdimsdale fixed the tests - only took 1.5 years.

williammartin:
Thanks. I really love the diff-ing, but the file writing and dependency on a system tool makes me slightly cautious. Are there any examples of either of these things in Gomega already?

raphael:
Totally understand being cautious here. A few considerations:

* `gexec` already use temporary files
* The code degrade gracefully if the `diff` tool is not present.
* The diff tool itself is a very stable dependency (at least on *nix systems)
* I have never run into issues using this and I use it a lot

At the end of the day this seems to be a case where the practicality of the feature overrides the conceptual issues with its implementation. That being said I'm not the one maintaining this package :)

williammartin:
We've been around and around on this pull request multiple times and we can never form consensus on it. 

While we recognise the value of a good diff-ing solution, we just can't get around the external dependency and the technical debt this could result in for different platforms, particularly if we ever tried to produce an alternative diff solution for said platforms. As such, I'm going to close this out and if there's outrage we can revisit, rather than dealing from the position of status quo where information about appetite has been lost over time.

As a total aside, I think any diff-ing solution we accepted would need to start off by default, at least until we fully understood what we wanted output to look like i.e. readable diff, avoiding redundant/repeated output. I'm not convinced by the output snippets here as the UX (it would frustrate me).

Thanks for submitting this, and I'm sorry it sat for so long, the pragmatism angle is appreciated, hence why we struggled so much.

PR #113, opened 2015-10-06
Reformat

glyn:
gomega seems to need some reformatting to match go fmt rules.

Signed-off-by: Glyn Normington <gnormington@pivotal.io>


onsi:
Thanks @glyn !

PR #117, opened 2015-12-09
Fix typo "iff" in documentation comment

bradylove:


jim-slattery-rs:
`iff` is probably not a typo: https://en.wikipedia.org/wiki/If_and_only_if

bradylove:
Ah very interesting, thank you. I think, based on how I interpret that, this use is correct. Will close the PR.

PR #118, opened 2015-12-09
fixed up: Succeed(), HaveOccurred()

jim-slattery-rs:
- Succeed() would panic if error was a nil pointer, like: `*CustomError`
- Succeed() and HaveOccurred() weren't both rejecting non-error types in all cases.

Added new tests for the above.


onsi:
Good stuff, thanks @jim-slattery-rs - I agree with your comment that there is shared behavior here that could be factored out but there's not a huge amount of duplication so I'm not too concerned.  Will pull this in as is.

PR #121, opened 2015-12-27
Add RespondWithCallback handler.

robdimsdale:
- useful for when the response needs to be dynamic.

### DIFF
--- a/ghttp/handlers.go
+++ b/ghttp/handlers.go
@@ -209,6 +209,22 @@ func RespondWith(statusCode int, body interface{}, optionalHeader ...http.Header
 }
 
 /*
+RespondWithCallback returns a handler that responds to a request with the specified status code and result of evaluating the provided callback.
+
+Also, RespondWithCallback can be given an optional http.Header.  The headers defined therein will be added to the response headers.
+*/
+func RespondWithCallback(statusCode int, callback func() []byte, optionalHeader ...http.Header) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if len(optionalHeader) == 1 {
+			copyHeader(optionalHeader[0], w.Header())
+		}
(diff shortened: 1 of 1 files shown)


onsi:
Hey @robdimsdale what if instead we update `RespondWith` to also accept a body of type `func() []byte` or `func() string`?  That would avoid a new top level function and keep everything in `RespondWith`

Thoughts?

onsi:
Also, do u mind updating the gh-pages branch with some documentation?  Thanks!

robdimsdale:
Sure I'll update the documentation. I think you're suggesting having `RespondWith(statusCode int, body interface{}, optionalHeader ...http.Header)` switch on `body` being a callback? I'll take a look and see if I can get that working.

Separately, as I play with this some more, I actually think I want the callback to take the same `req *http.Request` that the handler takes, allowing for dynamic behavior based on the properties of the request. I'll look into that too

robdimsdale:
I will also take a peek at the tests

onsi:
If that's the case then, since these are just `http.HandlerFuncs` why not simply pass in your own handler instead of wrapping things with `RespondWith`?

robdimsdale:
Because I completely forgot that was a thing that gingko supports! I was focussing on the input side of things (i.e. I wasn't particularly interested in the `http.Writer` side of the `http.Handler`, but only the `req *http.Request`).

Thanks for the suggestion. If I can get what I want without changing gomega's codebase I'll do that, otherwise I'll follow through on this conversation. Will update this thread shortly.

robdimsdale:
Your suggestion was spot-on, thanks! I can achieve what I want with a combination of `server.RouteToHandler` and custom handlers. I'll close this PR as it's not necessary.

onsi:
Great!  Glad that could help :)

PR #125, opened 2016-01-27
Allow the ghttp server to be reset

dprotaso:
There are cases were we want to start one http server in BeforeSuite and
reset its handlers & state between tests

Signed-off-by: Topher Bullock <cbullock@pivotal.io>


onsi:
Looks good, but I gotta ask:  Why not always create a new ghttp server in a `BeforeEach`?

PR #130, opened 2016-02-18
Use reflect's TrySend instead of reflect.Select.

apoydence:
Small refactor.


apoydence:
Oops. Looks like this doesn't work with older version of Go. Never mind then!

PR #132, opened 2016-02-20
Improves MatchJSON error message

tjarratt:
When one of the actual or expected cannot be parsed as JSON, it would be quite nice to know quickly which one you, kind developer, should be looking at in order to improve and fix your tests.

This commit mostly just adds the words "expected" and "actual" around the error messages, which is hopefully more clear.

@robdimsdale I tried using `json.Marshal` here but it actually yields back a string if you try to marshal a string, which isn't very helpful. My expectation is that `json.Marshal("ruh roh")` which return an error, but it really just yields back `"ruh roh"`.

Wrapping the error with some more precise language seemed like the correct thing to do here, but I'm open to any feedback.

A follow-up to this PR that I would like to do next is to check if the JSON matches, and if not, walk each structure recursively until one field or member of the JSON is found to differ. Is this what you had in mind when you said a "semantically aware, object-level diff", @robdimsdale?

### DIFF
--- a/matchers/match_json_matcher.go
+++ b/matchers/match_json_matcher.go
@@ -4,8 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/onsi/gomega/format"
 	"reflect"
+
+	"github.com/onsi/gomega/format"
 )
 
 type MatchJSONMatcher struct {
@@ -50,11 +51,11 @@ func (matcher *MatchJSONMatcher) prettyPrint(actual interface{}) (actualFormatte
 	ebuf := new(bytes.Buffer)
 
 	if err := json.Indent(abuf, []byte(actualString), "", "  "); err != nil {
-		return "", "", err
+		return "", "", fmt.Errorf("Actual '%s' should be valid JSON, but it is not.\nUnderlying error:%s", actualString, err)
 	}
 
 	if err := json.Indent(ebuf, []byte(expectedString), "", "  "); err != nil {
-		return "", "", err
+		return "", "", fmt.Errorf("Expected '%s' should be valid JSON, but it is not.\nUnderlying error:%s", expectedString, err)
 	}
 
 	return abuf.String(), ebuf.String(), nil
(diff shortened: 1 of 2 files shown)


robdimsdale:
Thanks for the PR! This all looks great to me. I was surprised at the fact `json.Marshal("ruh roh")` does not return an error, but I guess a raw string is valid json? I tried it out on the [go playground](https://play.golang.org/p/hXRRarHfoX) in case anyone's curious.

Although the tests are failing on go1.5, they're not failing for this PR so I'd be happy to merge if @onsi has no objections. I don't think there's any docs changes necessary for this PR.

As far as the more intelligent matching goes, I think we're on the same page. Recursively walking each element in the `map[string]interface{}` and comparing the value would be great. This would greatly improve the quality of the output and hence the usability of the `MatchJSON` method.

You may find writing the treewalker a little tedious, but I've seen examples of code that performs this and it doesn't look particularly hard. I would suggest the treewalker has its own set of unit tests, so we don't have to throw lots of different structures into the `MatchJSON` test.

As long as we don't change the interface of `MatchJSON` I don't think it'll be a breaking change, so you won't need a new matcher - just utilize the treewalker in the `MatchJSON` method.

Lastly for this proposed treewalking PR, I don't think there'll be any docs necessary, but it might be worth updating them so consumers know what to expect. I've heard more than one person say they wished gomega had this feature, and updating the docs helps spread the word.

tjarratt:
A raw string, on its own, is not valid JSON -- only objects and arrays are valid top-level JSON constructs -- it's just that `json.Marshal` documents that it "yields the JSON encoding of (the provided `interface{}`) value". Hence why it doesn't return an error, unless the value cannot possibly be represented as JSON. 

Apparently `json.Marshal` only returns errors if you try to encode channels, so-called "complex" values, or functions.

Thanks for the feedback @robdimsdale, I really appreciate it. Do you want to keep this PR open while I continue working on the matcher?

robdimsdale:
Thanks for the clarification on json behavior @tjarratt. I'd be happy to merge but I'd like to give @onsi a chance to take a look first.

I would probably build your next PR off this code, as we are likely to merge it, but I would open a separate PR for the next submission so we can merge this one without impacting future work.

tjarratt:
:+1:

onsi:
LGTM!

Issue #104, opened 2015-08-19
Comparing pointers?


elliotchance:
I want to test if two objects have the same pointer:

```go
func GetRoot() *BluePrintObject {
    return new(BluePrintObject)
}
```

```go
var _ = Describe("Blueprint", func() {
    Context("The root node", func () {
        var rootNode = GetRoot()
        It("should be a singleton", func () {
            var anotherRootNode = GetRoot()
            Ω(rootNode).Should(Equal(anotherRootNode))
        })
    })
})
```

This passes, however I want it to fail since they are different memory locations. How about a new matcher?

```go
Ω(rootNode).Should(BeTheSameObjectAs(anotherRootNode))
```

elliotchance:
For the time being I can use:

```go
Ω(rootNode == anotherRootNode).Should(BeTrue())
``

onsi:
hey @elliotchance I'd be happy to pull in a PR if you're up for putting it together.  Should be fairly straightforward to copy and paste your way to victory though you might want to make sure to handle `nil` appropriately (might just work, always a tricky edge case).

A code PR + docs PR to the gh-pages branch would be great.

As for the matcher name, another option could be `BeIdenticalTo()` or `StrictlyEqual()`.

tjarratt:
:+1: would love to have a `BeIdenticalTo()` matcher in Gomega. This is probably one of my favorite matchers from Cedar that I've been missing when writing Go.

@elliotchance -- were you working on a PR? Would you be terribly upset if I worked on this?

elliotchance:
@tjarratt upset from someone helping? Never! :) Go for it.

Sorry @onsi, I forgot entirely about this request.

onsi:
Thanks  @tjarratt

tjarratt:
Thank you @onsi and @elliotchance!

PR #133, opened 2016-02-22
Add BeIdenticalTo matcher

tjarratt:
This aims to address issue #104. This should be quite similar to the `BeSameInstanceAs` matcher from Cedar, one of my personal favorites.

Internally it uses the == operator, and should only positively match primitive values as well as identical pointers to a value.


onsi:
Great!

Issue #131, opened 2016-02-19
MatchJSON failure messages are opaque and (at times) misleading


tjarratt:
I was using the `MatchJSON` matcher recently and had a heck of a time figuring out what some of the error messages bubbled up from the `encoding/json` package actually meant.

In particular, when neither the `actual`, nor the `expected` is valid as JSON, it's very hard to tell what to change next -- most of the time I had to resort to printing out both and trying to parse it as JSON myself.

It would be really nice if the error messages could be improved to indicate which of the two cannot be parsed as JSON which this occurs.

robdimsdale:
Agreed. I think a PR to this section would be well received. A simple change might be to add `json.Marshal` on both the expected and the actualy to https://github.com/onsi/gomega/blob/master/matchers/match_json_matcher.go#L48 

A much more ambitious solution would be to implement a more semantically-aware matcher that wouldn't just dump the string contents, but actually provide a meaningful object-level diff.

tjarratt:
Agreed 100%

I was thinking of implementing the small PR and then tackling the larger
issue of improving how the matcher compares the json structures later.

If someone else wants to contribute a PR for this, I wouldn't be at all
upset. This seemed like a nice feature for a new contributor (such as
myself).

On Friday, February 19, 2016, Rob Dimsdale <notifications@github.com> wrote:

> Agreed. I think a PR to this section would be well received. A simple
> change might be to add json.Marshal on both the expected and the actualy
> to
> https://github.com/onsi/gomega/blob/master/matchers/match_json_matcher.go#L48
>
> A much more ambitious solution would be to implement a more
> semantically-aware matcher that wouldn't just dump the string contents, but
> actually provide a meaningful object-level diff.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/onsi/gomega/issues/131#issuecomment-186497475>.
>


-- 
Magnetized From My Etch-A-Sketch

robdimsdale:
A PR would be great. Do you feel like you have enough guidance to tackle this, or can I help further?

tjarratt:
I feel comfortable tackling this.

Thanks for the very kind words and support, Rob.

On Friday, February 19, 2016, Rob Dimsdale <notifications@github.com> wrote:

> A PR would be great. Do you feel like you have enough guidance to tackle
> this, or can I help further?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/onsi/gomega/issues/131#issuecomment-186499941>.
>


[...]

PR #136, opened 2016-03-02
Make the error message for expected JSON values having the wrong type…

tinygrasshopper:
… accurate

Before the fix we had a scenario in which

```go
someJSON := `{"description": "this is the actual"}`
Expect(someJSON).To(MatchJSON(map[string]string{"description": "this is the expected"}))
```

Used to fail with the output:

```
      MatchJSONMatcher matcher requires a string or stringer.  Got:
          <string>: {"description": "this is the actual"}
```

This shows the actual, when it is the expected that is of the wrong type.

After the fix, this fails with the output:

```
      MatchJSONMatcher matcher requires a string, stringer, or []byte.  Got expected:
          <map[string]string | len:1>: {
              "description": "this is the expected",
          }
```

Signed-off-by: Tom Anderson <toanderson@pivotal.io>

### DIFF
--- a/matchers/match_json_matcher.go
+++ b/matchers/match_json_matcher.go
@@ -40,11 +40,13 @@ func (matcher *MatchJSONMatcher) NegatedFailureMessage(actual interface{}) (mess
 }
 
 func (matcher *MatchJSONMatcher) prettyPrint(actual interface{}) (actualFormatted, expectedFormatted string, err error) {
-	actualString, aok := toString(actual)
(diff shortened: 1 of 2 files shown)


tinygrasshopper:
Actually this PR solves the problem stated in #131, as well

robdimsdale:
@tinygrasshopper @tomwhoiscontrary thanks for this. The tests passed locally so I'll merge.

PR #138, opened 2016-03-09
Fixes Doc Rendering 

tinygrasshopper:
same fix as onsi/ginkgo#233


robdimsdale:
Thanks for doing this (and https://github.com/onsi/ginkgo/pull/233) - it's much appreciated. Looks great, so I'll merge now.

Issue #139, opened 2016-03-16
YAML matcher


craigfurman:
Hi Onsi,

@tomwhoiscontrary and I created a MatchYAML matcher, which is heavily inspired by MatchJSON. It's in a [private CF repo that you should have access to](https://github.com/pivotal-cf-experimental/kafka-example-service-adapter/blob/master/yamlmatcher/match_yaml_matcher.go). It has a dependency on gopkg.in/yaml.v2.

Would you be interested in including this in Gomega? If so we can PR it in.

robdimsdale:
@craigfurman @tomwhoiscontrary thanks for doing this! I know it'll be useful.

There is some pain around ginkgo/gomega pulling in dependencies, see #114 and #123, so I'm not sure adding another dependency is a path we want to go down.

If you are willing to extract the matcher into its own (publically-available) repo, update the gomega docs to point to it, and add me as a maintainer, I would be happy to help out.

Thanks for doing this,
Rob

craigfurman:
Sounds good. We've created a chore
<https://www.pivotaltracker.com/story/show/115919467> in our private
tracker to do this

tinygrasshopper:
@robdimsdale It seems a bit odd to me to have a separate repo with one matcher.

Suggestion, could we have a separate package for the matchers requiring external dependencies.
It might look like:
`gomega/matchers/extensions/yaml`
`gomega/matchers/extensions/protobuf`

The difference between the extended matchers and the normal matchers, they wont be included in the `gomega/matchers.go`, so they wont be fetched when the user invokes `go get github.com/onsi/gomega`. If someone would like to use one of the extended matchers, they would have to fetch it by `go get gomega/matchers/extensions/yaml`

On the same lines as `ginkgo/extensions/table`.

What do you think?

robdimsdale:
@tinygrasshopper I think that proposal also works. Essentially I think the issue I'm looking to avoid is having 3rd-party dependencies imported from the root package, either directly or transitively, as this requires `go get github.com/onsi/gomega/...` rather than simply `go get github.com/onsi/gomega`.

I agree that splitting matchers across multiple repos is no ideal as it increases the cost of discovery. Your suggestion seems to be a good compromise this and the problem of introducing a new dependency.

We can't move the `protobuf` without making a breaking change (see #123) but we have the freedom to choose where new dependencies go. I like the idea of leveraging the `extensions` package like we do with `ginkgo`.

@onsi what are your thoughts?

[...]

PR #141, opened 2016-03-24
Gomega yaml matcher

mariantalla:
As discussed in #139 , we are submitting the code for the gomega YAML matcher.


robdimsdale:
I think this looks ok, but I'd like to give @onsi the opportunity to chime in on the topic of using the `extensions` package for matchers with third-party dependencies?

Given a similar situation in https://github.com/onsi/ginkgo/pull/239 and the fact we have already established a pattern in this repo for [copying in dependencies like goraph](https://github.com/onsi/gomega/tree/master/matchers/support/goraph) we may yet decide to go down an alternate route.

tinygrasshopper:
@onsi any thoughts on this PR?

rosenhouse:
@robdimsdale @tinygrasshopper  Surprised to see this merged:

- The go-yaml dependency has a non-standard license, which has unresolved legal issues for Pivotal.

- The Release Infrastructure team built [their own yaml matcher](https://github.com/pivotal-cf-experimental/gomegamatchers) but was under the impression at the time that @onsi wasn't interested in getting a PR for gomega, due to the extra required dependency.

cc: @Amit-PivotalLabs @zankich

robdimsdale:
Sorry for the confusion here.

I can't speak to the first issue, but we've been using that library throughout Pivotal projects for months - personally I was under the impression it's either been cleared, or else it's not a problem.

As to the second, we didn't want to merge a matcher because we (incorrectly) thought it would cause more breakages along the lines of the protobuf transitive dependencies - this was recently disproved by the london team.

If we have a better yaml matcher we should look at improving this one - either way, consolidation would be great.

PR #142, opened 2016-03-24
matching docs for YAML matcher

mariantalla:
Docs for #141 Ref: #139


robdimsdale:
@mariantalla thanks for this PR. These docs look fine, and assuming they are still relevant when https://github.com/onsi/gomega/pull/142 is resolved we can merge this.

PR #145, opened 2016-04-16
Reset tmpDir in gexec.CleanupBuildArtifacts

st3v:
If `tmpDir` is not being reset in `gexec.CleanupBuildArtifacts`, subsequent calls to `gexec.Build` are failing with an `os.PathError`. This is because `gexec.temporaryDirectory` won't recreate the `gexec_artifacts` temp dir if `tmpDir` is set. This is a problem when `Build` and `CleanupBuildArtifacts` are being used in `BeforeEach` and `AfterEach` blocks that are being executed multiple times.

### DIFF
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -62,6 +62,7 @@ gexec. In Ginkgo this is typically done in an AfterSuite callback.
 func CleanupBuildArtifacts() {
 	if tmpDir != "" {
 		os.RemoveAll(tmpDir)
+		tmpDir = ""
 	}
 }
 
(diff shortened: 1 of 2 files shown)


robdimsdale:
Thanks @st3v - this looks good to me. I don't think the travis test failures are related.

I can't see any reason why we shouldn't support `CleanupBuildArtifacts` in `BeforeEach` vs `BeforeSuite` (which is the more common pattern), so I would be happy to merge this unless @onsi has any objections.

svett:
@onsi shall we merge this? All tests are passing.

onsi:
👍🏽

svett:
Thanks @st3v. I merged your changes.

PR #148, opened 2016-04-24
Always escape a slice of bytes

### LINKED ISSUE #147: Always escape a slice of bytes
The type `[]byte` should be seen as memory and is mostly used for buffers and thus contain non-printable characters most of the time. Some of these non-printable characters can even be interpreted by the terminal, leading to bad side-effects (emacs' compilation buffer segfaulting in my case...).

I extracted the underlying formatting function gomage uses for `[]byte` to show you an example of what happens: https://play.golang.org/p/ycVWoncyYy

Related: https://blog.golang.org/strings

Julio-Guerra:
Don't use `formatString()` to format a value of type `[]byte` but
`formatSlice()` which formats the slice into an array of `uint8`.

Fixes #147

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -143,9 +143,6 @@ func formatValue(value reflect.Value, indentation uint) string {
 	case reflect.Ptr:
 		return formatValue(value.Elem(), indentation)
 	case reflect.Slice:
-		if value.Type().Elem().Kind() == reflect.Uint8 {
-			return formatString(value.Bytes(), indentation)
-		}
 		return formatSlice(value, indentation)
 	case reflect.String:
 		return formatString(value.String(), indentation)
--- a/format/format.go
+++ b/format/format.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"strconv"
 )
 
 // Use MaxDepth to set the maximum recursion depth when printing deeply nested objects
@@ -186,6 +187,10 @@ func formatString(object interface{}, indentation uint) string {
 }
 
 func formatSlice(v reflect.Value, indentation uint) string {
+	if v.Kind() == reflect.Slice && v.Type().Elem().Kind() == reflect.Uint8 && isPrintableString(string(v.Bytes())){
+		return formatString(v.Bytes(), indentation)
+	}
+
 	l := v.Len()
 	result := make([]string, l)
 	longest := 0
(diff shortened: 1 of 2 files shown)


robdimsdale:
@Julio-Guerra thanks for this. Is there a way you could add an accompanying test for this? Also, it seems like this change caused some new test failures: https://travis-ci.org/onsi/gomega/jobs/125337691

Julio-Guerra:
Oops. I'm working on it. Existing (broken) tests already cover this.

svett:
@robdimsdale @onsi: There is a test coverage for this change. Shall I merge? Everything passes.

Julio-Guerra:
ping

onsi:
Sorry @Julio-Guerra - just getting to this :/

Can you share why you prefer this approach?  In my experience []byte arrays in test suites usually (granted: not always, but usually) are encoding strings.  Seeing strings rendered in the failure messages is more helpful than seeing the byte array itself.

I can imagine a compromise where *both* representations are rendered... but that might be confusing and/or verbose.

Wonder what others on the issue think.

xoebus:
What if we check if all the characters in the actual in expected strings are valid code points and if so, output as a string, otherwise show the byte array?

Julio-Guerra:
I explained why in issue #147. I am generating random buffers to test a pty
driver and they very often include special terminal sequences. Emacs was
even segfaulting some times. I don't know if some escape sequences could be
unexpected and thus lead to undefined behaviours...

As @xoebus said, we could check for non-printable character, having to go
through every single byte in the slice in the worst case to be able to
print as a string.
These `[]byte` strings could no longer be printed with terminal escape
sequences.


Le jeu. 18 août 2016 07:14, Christopher Brown <notifications@github.com> a
écrit :

onsi:
I suspect we don't need to worry too much about performance.  are you up
for updating the PR?

Julio-Guerra:
Done.
Since CI problems were fixed on the master branch and this PR branch was old, I rebased it and had to do a `push -f` it, this is why you see these 2 old commits reappear.

I reverted the tests to get back the string tests against byte slices and added a specific new one including non-printable characters (according to `strconv.IsPrint()`).

jvshahid:
I personally like seeing the string diff. :+1: for doing a string diff and falling back to byte slice diff if the byte slice has invalid code points

onsi:
thanks!

PR #150, opened 2016-05-10
Fixing build on travis

tinygrasshopper:
Not re-using of TCP connections between http requests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
   - 1.5
-  - tip
+  - stable
 
 install:
   - go get -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 go:
   - 1.5
   - 1.6
+  - 1.6.2
   - stable
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 go:
   - 1.5
-  - 1.6
   - 1.6.2
   - stable
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 go:
   - 1.5
+  - 1.6
   - stable
 
 install:
--- a/ghttp/test_server_test.go
+++ b/ghttp/test_server_test.go
@@ -55,8 +55,8 @@ var _ = Describe("TestServer", func() {
 					io.WriteString(w, req.RemoteAddr)
 				},
 			)
-
-			resp, err := http.Get(s.URL())
+			client := http.Client{Transport: &http.Transport{DisableKeepAlives: true}}
+			resp, err := client.Get(s.URL())
 			Ω(err).ShouldNot(HaveOccurred())
 			Ω(resp.StatusCode).Should(Equal(200))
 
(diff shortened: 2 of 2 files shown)


robdimsdale:
Thanks @tinygrasshopper 

I ran these tests locally (go v1.6.2) with:

```
ginkgo -p -r . -untilItFails
```

What I saw was that we got about three successful runs and then an unrelated test failed on the fourth run. However, when I check out the previous SHA and run the same command, I also see all tests pass - I'm currently on test run eleven and haven't seen any failures yet.

Which tests were you looking to fix here?

Thanks,
Rob

tinygrasshopper:
The error that I was trying to fix is the error related to reusing of connection pool, which can be seen [here](https://travis-ci.org/onsi/gomega/jobs/113955433).

I did manage to fix it on [v1.5](https://travis-ci.org/onsi/gomega/jobs/129059266) and v1.6 

But on go tip, we are getting a [data race](https://travis-ci.org/onsi/gomega/jobs/129059269) which seems unrelated, but I cant quite understand why.

I propose testing on the last stable version of go rather than the go tip, which is the HEAD of master at golang/go

tinygrasshopper:
`stable` on travis is extremely outdated, currently its failing on v1.6 on a timing issue. 

My overall goal here was to get a green build on travis for the project.

tinygrasshopper:
I should have been more descriptive on what I was trying to do with this PR.

Goal: Getting the build passing on travis

Changes
- Use a disable keep alive when testing closing client connection in ghttp, as it was causing failures like [this](https://travis-ci.org/onsi/gomega/jobs/113955433)
- Stop testing against go tip, and test against the last stable version, 1.6.2. On the current go tip the build fails due to a data race involving [sync/atomic](https://travis-ci.org/onsi/gomega/jobs/129059269), which is strange . Do we want to test ginkgo against the latest commit of go, I think we should test against the latest stable version.

Any thoughts @robdimsdale @onsi?

tinygrasshopper:
@robdimsdale @onsi any thoughts on this? It is useful to see if the build is passing for a new PR on travis, so that we don't have to run it on local before merging. The only downside is we don't test against the latest commit on golang/go#master, which I think is ok.

robdimsdale:
Firstly, thanks for explaining your goals and detailing the consequences of the changeset - that's really helpful.

I think broken tests now are a worse situation than potentially broken tests in the future, so I'd like to merge this to fix the currently-broken tests.

As far as the changes go, the client change seems fine to me. In the travis yml file, do we need to explicitly test `go: 1.6` - isn't `go 1.6.2` sufficient?

If possible, it would be useful to test against go-tip so we can be ahead of the changes that are coming - if there are consistent failures it could prove to be a useful canary to tell us that the next minor release of go might be incompatible with ginkgo/gomega

That being said, I'd like to see the `1.6.2` tests passing as it's a pretty-big [broken-window](https://en.wikipedia.org/wiki/Broken_windows_theory) - especially for PRs from the community.

I'd love it if there were a way it didn't break on `tip`, but if I could only pick one of `tip` and `1.6.2`, I'd choose `1.6.2`.

tinygrasshopper:
I have removed `go: 1.6` from travis.yml

I agree that it would be good to know if the tests are passing against `tip`, but the way I see it `tip` is a moving target, and changes nearly everyday. If we are seeing failures on tip which are not easily identifiable I don't think we should look into them, till tip reaches stable.

I think this PR can be merged in as it does give us feedback for future PR's against 1.6.2. What do you think? We will have to come up with a way to check against tip as well, but I can't think of a good one right now.

robdimsdale:
Looks good, thanks. I agree with your thoughts on tip, and I'm happy to defer the question of testing future versions of golang until we observe it becoming an issue.

@onsi any objections to merging this?

PR #151, opened 2016-05-12
Ambiguous formatting of numbers 

tinygrasshopper:
Formatting of floats and ints is ambiguous, floats and ints are displayed the same way in the assertion output.

For example, for this test:

```
    FIt("fails", func() {
        a := map[string]interface{}{
            "a": 1,
        }
        data, _ := json.Marshal(a)
        var v map[string]interface{}
        json.Unmarshal(data, &v)

        Expect(v).To(Equal(a))
    })
```

Currently:

```
      Expected
          <map[string]interface {} | len:1>: {"a": 1}
      to equal
          <map[string]interface {} | len:1>: {"a": 1}
```

After the fix:

```
      Expected
          <map[string]interface {} | len:1>: {"a": 1.0}
      to equal
          <map[string]interface {} | len:1>: {"a": 1}
```

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -5,6 +5,7 @@ package format
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"strings"
 )
@@ -133,7 +134,11 @@ func formatValue(value reflect.Value, indentation uint) string {
 	case reflect.Uintptr:
 		return fmt.Sprintf("0x%x", value.Uint())
 	case reflect.Float32, reflect.Float64:
-		return fmt.Sprintf("%f", value.Float())
+		if math.Trunc(value.Float()) == value.Float() {
+			return fmt.Sprintf("%0.1f", value.Float())
+		} else {
+			return fmt.Sprintf("%v", value.Float())
+		}
 	case reflect.Complex64, reflect.Complex128:
 		return fmt.Sprintf("%v", value.Complex())
 	case reflect.Chan:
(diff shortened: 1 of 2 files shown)


format/format.go
```diff
@@ -133,7 +133,7 @@ func formatValue(value reflect.Value, indentation uint) string {
 	case reflect.Uintptr:
 		return fmt.Sprintf("0x%x", value.Uint())
 	case reflect.Float32, reflect.Float64:
-		return fmt.Sprintf("%v", value.Float())
+		return fmt.Sprintf("%f", value.Float())
```
jim-slattery-rs: I believe this limits the output to 6 decimal places. Maybe instead use `strconv.FormatFloat` with precision set to -1?
https://golang.org/pkg/strconv/#FormatFloat

tinygrasshopper: Interesting, didn't realize that. I guess thats why it was %v to begin with.

I guess there should be separate specs for floats with decimal places and the one without

jamesjoshuahill:
🐬 for fixing this matcher output

johanbrandhorst:
This would be a great addition :+1:

williammartin:
@tinygrasshopper Do you still care about this PR? I see the remote branch doesn't exist anymore.

tinygrasshopper:
I am ok to close it. I can open a new one if I run into this issue again

PR #152, opened 2016-05-20
Tracking gexec sessions

### LINKED ISSUE #52: Gexec should track all running processes and provide a mechanism to clean them all up
e.g.
```
gexec.Interrupt([timeout])
gexec.Kill([timeout])
gexec.Signal(signal, [timeout])
```

tinygrasshopper:
Providing a mechanism to track and send signals to all the processes started using `Start`. Useful to cleanup processes after the tests, to prevent processes going rogue.

Added the methods:
`Kill`
`KillAndWait`
`Terminate `
`TerminateAndWait`
`Signal `
`Interrupt `
to send signals to all the processes tracked.

resolves #52

### DIFF
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -92,6 +92,8 @@ func Start(command *exec.Cmd, outWriter io.Writer, errWriter io.Writer) (*Sessio
 	err := command.Start()
 	if err == nil {
 		go session.monitorForExit(exited)
+		trackedSessionsMutex.Lock()
+		defer trackedSessionsMutex.Unlock()
 		trackedSessions = append(trackedSessions, session)
 	}
 
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -92,9 +92,9 @@ func Start(command *exec.Cmd, outWriter io.Writer, errWriter io.Writer) (*Sessio
 	err := command.Start()
 	if err == nil {
 		go session.monitorForExit(exited)
+		trackedSessions = append(trackedSessions, session)
 	}
 
-	trackedSessions = append(trackedSessions, session)
 	return session, err
 }
 
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -215,13 +215,7 @@ func (s *Session) monitorForExit(exited chan<- struct{}) {
 }
 
 var trackedSessions = []*Session{}
-
-/*
-Resets the sessions tracked by gexec after Run is called.
-*/
-func ResetTrackedSessions() {
-	trackedSessions = []*Session{}
-}
+var trackedSessionsMutex = &sync.Mutex{}
 
 /*
 Kill sends a SIGKILL signal to all the processes started by Run, and waits for them to exit.
@@ -230,6 +224,8 @@ The timeout specified is applied to each process killed.
 If any of the processes already exited, KillAndWait returns silently.
 */
 func KillAndWait(timeout ...interface{}) {
+	trackedSessionsMutex.Lock()
+	defer trackedSessionsMutex.Unlock()
 	for _, session := range trackedSessions {
 		session.Kill().Wait(timeout...)
 	}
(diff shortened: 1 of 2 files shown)


onsi:
Looks good, couple thoughts:

- I think we want a lock around all access to `trackedSessions`
- Would be good to update the gomega docs and describe best practices (e.g. calling `KillAndWait()` in a top-level `AfterSuite` - given the global nature of this addition it could be an error to do it in an `AfterEach` since processes started in `BeforeSuite` would get clobbered).
- When would a user use `ResetTrackedSessions()` - I'm just wondering what the best practice will look like.  Ideally (though it would be hard) you could imagine `trackedSessions` getting cleaned up automatically whenever a process exits.  Since the point of the feature is to help users totally clean up their sessions I wonder if it would best if we *didn't* have `ResetTrackedSession()` but instead encouraged folks to always call `KillAndWait()` to clean things up?

tinygrasshopper:
Thanks for looking into this,
- Agreed about the lock, totally missed that one. 
- Will enhance the docs as well. 
- `ResetTrackedSessions` was only really written to allow to me test drive the functionality. Now that I think about it I can't think of any scenarios in which I would like the end-user to actually use it. The users should always use the other functions as they are no-ops when the process has already died. Will change the test cleanup to `KillAndWait ` rather than `ResetTrackedSession`, I think it should be equivalent.

onsi:
Gah! sorry, this fell off my radar.  looks like travis is failing on 1.5 - and it looks like its related to the new code?

tinygrasshopper:
@onsi oh didn't notice that. Will take a look. 

On a similar note, what are your thoughts on merging #150, so we can get rid of the false negatives from the PR build status?

onsi:
fine by  me!

tinygrasshopper:
@onsi Fixed the travis failures, thoughts?

gexec/session.go
```diff
@@ -92,6 +92,7 @@ func Start(command *exec.Cmd, outWriter io.Writer, errWriter io.Writer) (*Sessio
 	err := command.Start()
 	if err == nil {
 		go session.monitorForExit(exited)
+		trackedSessions = append(trackedSessions, session)
```
onsi: @tinygrasshopper you're missing a lock here

gexec/session.go
```diff
@@ -212,3 +213,91 @@ func (s *Session) monitorForExit(exited chan<- struct{}) {
...
+	}
+}
+
+/*
+Kill sends a SIGKILL signal to all the processes started by Run.
+It does not wait for the processes to exit.
+
+If any of the processes already exited, Kill returns silently.
+*/
+func Kill() {
```
onsi: This could use `Signal` under the hood, no?  bit less code duplication?

gexec/session.go
```diff
@@ -212,3 +213,91 @@ func (s *Session) monitorForExit(exited chan<- struct{}) {
...
+	}
+}
+
+/*
+Terminate sends a SIGTERM signal to all the processes started by Run.
+It does not wait for the processes to exit.
+
+If any of the processes already exited, Terminate returns silently.
+*/
+func Terminate() {
```
onsi: This could use Signal under the hood, no? bit less code duplication?

gexec/session.go
```diff
@@ -212,3 +213,91 @@ func (s *Session) monitorForExit(exited chan<- struct{}) {
...
+
+If any of the processes already exited, KillAndWait returns silently.
+*/
+func KillAndWait(timeout ...interface{}) {
+	trackedSessionsMutex.Lock()
+	defer trackedSessionsMutex.Unlock()
+	for _, session := range trackedSessions {
+		session.Kill().Wait(timeout...)
+	}
+	trackedSessions = []*Session{}
```
onsi: This is the only method that clears the `trackedSessions` - is that by design?  Or should all the others also include this line?

gexec/session.go
tinygrasshopper: Fixed, good catch

gexec/session.go
tinygrasshopper: After kill and wait is the only time we know for sure that the process has actually exited. The other times the process might choose to ignore the signal.

onsi:
pulling it in!

PR #159, opened 2016-06-24
Allow gexec.Build to be called concurrently

WesleyJeanette:
This will allow us to build multiple binaries concurrently.


onsi:
durp.  Thanks y'all.

PR #163, opened 2016-07-10
Remove dead func

jasonkeene:
Pretty sure this isn't being used.


robdimsdale:
I agree; looks unused. Merged, thanks!

PR #164, opened 2016-07-12
Fix typo: asser -> assert

wendorf:


robdimsdale:
Great, thanks for catching this!

PR #176, opened 2016-10-20
Print panic object when using ShouldNot(Panic())

petergtz:
When debugging the case where a `Ω(func() { ... }).ShouldNot(Panic())` does actually panic, it's often difficult to understand what happened. This PR prints the objects that was passed to `panic` and makes the debugging process easier.

### DIFF
--- a/matchers/panic_matcher.go
+++ b/matchers/panic_matcher.go
@@ -42,5 +42,5 @@ func (matcher *PanicMatcher) FailureMessage(actual interface{}) (message string)
 }
 
 func (matcher *PanicMatcher) NegatedFailureMessage(actual interface{}) (message string) {
-	return format.Message(actual, fmt.Sprintf("not to panic, but panicked with <%T>: %v", matcher.object, matcher.object))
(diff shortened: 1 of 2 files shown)


onsi:
good idea @petergtz - rather than use `%T and %v` do you mind resubmitting the PR using Gomega's `format.Object()` method.  This will recurse through the panic object if necessary and will make the output consistent across other matchers.

petergtz:
@onsi Makes sense.

Updated the PR. 3 questions/comments:

1. Do you prefer a squashed commit instead of the individual ones?
2. I've introduced a line break + indentation to follow output from other matchers. Let me know if you prefer to have it without the line break.
3. While doing this, I noticed that format.Object() formats string objects differently depending on the indentation: With no indentation it is `<string>: "my string"`; with indentation it is `<string>: my string`. Is this on purpose?

onsi:
1. individual is fine
2. line break looks good
3. hmm... i don't *think* that was intentional.

PR #179, opened 2016-10-26
Suppress printing the internals of context.Context. Added a flag to o…

ebabani:
…verride this behaviour.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -4,6 +4,7 @@ Gomega's format package pretty-prints objects.  It explores input objects recurs
 package format
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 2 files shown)


format/format.go
```diff
@@ -124,6 +135,12 @@ func formatValue(value reflect.Value, indentation uint) string {
 		}
 	}
 
+	if !PrintContextObjects {
+		if value.Type().Implements(contextType) && indentation > 1 {
+			return "<suppressed>"
```
onsi: perhaps this should be <suppressed context>?  wdyt?

onsi:
thanks @ebabani - i made a small comment on the commit.  can you also update the docs on gh-pages to mention this flag?  the relevant section is here:

http://onsi.github.io/gomega/#adjusting-output

ebabani:
@onsi I forgot that context.Context was added in 1.7. Can probably create our own interface to keep backwards compatability.

onsi:
ah good point.  yeah we'll need to do something to maintain backwards compatibility.... can you give it a shot?

format/format.go
ebabani: What do you mean here?

onsi: oh weird, github sanitized my comment.. `"<suppressed context>"`

ebabani:
Yeah, looks doable. The Context interface doesn't contain any 1.7 specific inputs or returns values.

PR #181, opened 2016-10-27
Add gexec.BuildWithEnvironment

### LINKED ISSUE #177: Use gexec.Build() to target other platforms?
@tinygrasshopper and I have been using `gexec.Build` to build binaries and test them, but sometimes we want to e.g. build a Linux binary, scp it to a remote machine and run it there.

Using `os.Setenv` to set the right env vars for targeting Linux seems a bit awkward. Thoughts on a better way of doing this? I'm happy to PR in a new `gexec.BuildForPlatform` method if there's interest.

Henry

henryaj:
Accepts a map of env vars as key-value pairs which are set during compilation. Useful for building a binary that gets tested on a remote VM/in a container that uses a different architecture to your local machine.

Resolves #177.

### DIFF
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -30,7 +30,7 @@ func Build(packagePath string, args ...string) (compiledPath string, err error) 
 /*
 BuildWithEnvironment is identical to Build but allows you to specify env vars to be set at build time.
 */
-func BuildWithEnvironment(packagePath string, env map[string]string, args ...string) (compiledPath string, err error) {
+func BuildWithEnvironment(packagePath string, env []string, args ...string) (compiledPath string, err error) {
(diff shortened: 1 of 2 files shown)


onsi:
hey @henryaj two asks:

1. Can you add a corresponding documentation PR to the `gh-pages` branch?
2. `map[string]string` makes a lot of sense but should this be `[]string` to match what's on the [`exec.Cmd`](https://golang.org/pkg/os/exec/#Cmd) object?

henryaj:
Sure thing @onsi -- now uses `[]string` args. Docs PR is [here](https://github.com/onsi/gomega/pull/183).

onsi:
LGTM!

PR #190, opened 2017-01-13
Attempt to diff long strings when not equal

ericTsiliacos:
Our goal in this PR was to improve the UX for `Equal()` when both of the expected and actual are very, very long strings.

Long strings (with more than 50 characters) that are not equal are typically
hard to visually diff when your tests fail. This change attempts to find the
very first mismatch in the strings and to visually point out where the diff
is. It additionally truncates the text, so it is easier to see the exact point.

Assumptions:
* people only need ~5 characters around the diff to identify WHERE it is
* 50 characters is a reasonable threshold for deciding to diff strings in this manner
* Equal() is the only matcher that needs this treatment (for now)

Would love feedback on these assumptions, or anything else.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -70,6 +70,74 @@ func Message(actual interface{}, message string, expected ...interface{}) string
 }
 
 /*
+
+Generates a nicely formatted matcher success / failure message
+
+Much like Message(...), but it attempts to pretty print diffs in strings
+
+Expected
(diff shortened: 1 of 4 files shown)


onsi:
hot.  can i come upstairs and take a look?

ericTsiliacos:
Yeah! We're in the beach corner near HR. Come on down!

PR #209, opened 2017-04-29
Adds support for MatchXML

aubm:
Actual and expected can not be pretty printed in the test output because
there are issues in the encoding/xml package related to deplicated
namespaces. See https://github.com/golang/go/issues/7535

Because Go has no support for collecting all XML attributes before 1.8
(see [here](https://github.com/golang/go/commit/c1a1328c5f004c62b8c08faf0d0d2845e0be5d37)), the two following XMLs will be equal for 1.7 and before:

```
<person gender="female">
```

```
<person gender="male">
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 go:
   - 1.6
   - 1.7
+  - 1.8
 
 install:
   - go get -v ./...
--- a/matchers.go
+++ b/matchers.go
@@ -214,6 +214,20 @@ func MatchJSON(json interface{}) types.GomegaMatcher {
 	}
 }
 
+//MatchXML succeeds if actual is a string or stringer of XML that matches
+//the expected XML.  The XMLs are decoded and the resulting objects are compared via
+//reflect.DeepEqual so things like key-ordering and whitespace shouldn't matter.
+//The comparison of attribute values is not supported for go1.7 and before.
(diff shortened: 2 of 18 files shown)


onsi:
Sweet!  Unfortunately it looks like this broke 1.6 and 1.7 on Travis. Any chance you could take a look?

aubm:
Yes I'm trying to figure it out.
I'm running out of time right now but I will come back to it later :)

onsi:
👍

aubm:
It seems that this particular struct field tag ``xml:",any,attr"`` was not supported before `1.8`. See https://github.com/golang/go/commit/c1a1328c5f004c62b8c08faf0d0d2845e0be5d37

I will have to search for another approach, too bad...

aubm:
Possibly there could be specific code for `<1.8`, minus the support for attributes checking.
It is not super clean... what is your opinion on this?

onsi:
I'm no expert on the vagaries of XML support in Go but I'm OK with version dependent behavior as long as it is documented clearly in godoc and the gh-pages branch.

aubm:
It should be ok now.

onsi:
Ok thanks.  Can you submit a PR for the gh-pages docs as well?

aubm:
Done: https://github.com/onsi/gomega/pull/210

onsi:
Sweet thanks!

PR #215, opened 2017-05-26
Make assertion failures be Fatal

jim-slattery-rs:
When using Gomega with Ginkgo, assertion failures are fatal and halt the current test.
But when Gomega is used by itself in Golang style XUnit tests, assertion failures
do not halt the current test.

This was because t.Errorf was called instead of t.Fatalf.

Fixed to call t.Fatalf instead.


onsi:
Yes that is closer to the intent - thanks!

PR #218, opened 2017-06-08
Get GOPATH from go internals

alex-slynko:
Starting Go 1.8 GOPATH is not required and has default value to `$HOME/go`.
https://golang.org/doc/go1.8#gopath
This change supports this behavior.

### DIFF
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -22,7 +22,8 @@ var (
 Build uses go build to compile the package at packagePath.  The resulting binary is saved off in a temporary directory.
 A path pointing to this binary is returned.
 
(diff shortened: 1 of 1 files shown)


williammartin:
@alex-slynko I'm just looking some PRs in flex time, this looks like it would put a dependency on Go 1.8+? Is there anywhere else in Gomega that enforces that kind of requirement? Also the comment:

```
Build uses the $GOPATH set in your environment.  It passes the variadic args on to `go build`.
```

would be incorrect.

alex-slynko:
1. There is no dependency on Go 1.8. Tests are passing on Go 1.6.
2. Comment is partially incorrect. Build uses $GOPATH set in the environment, for Go 1.8 it can use default environment.

PR #219, opened 2017-06-20
Added a matcher for json to allow for unordered lists

Benjamintf1:
Often you want to match against json but you don't know, or care what order a particular list is in. In this case, you might rely on MatchUnorderedJson. It treats all the lists in the son as unordered. If you want to have ordered keys, you can exclude them by using the OrderedKeys(matcher, keys) function to exclude some son keys from it's un-ordered grasp. 

As far as the tests, I just copied the regular json matcher tests(As they all should work) as well as some unordered tests(as they should also work). If you like this, I could add a "UnorderedKeys" function that operates on the MatchJson function to allow for some lists to be un-ordered. 

A co-worker suggested I make the code more "readable" when use, that's the reasoning behind using OrderedKeys(matcher, keys) instead of matcher(json, keys). 

If you like this, I can add the functionality/refactor this to matchXML and matchYML. 

Does any of this sound like a good idea/something you want in the base library?

### DIFF
--- a/matchers.go
+++ b/matchers.go
@@ -214,6 +214,16 @@ func MatchJSON(json interface{}) types.GomegaMatcher {
 	}
 }
 
+//MatchUnorderedJSON succeeds if actual is a string or stringer of JSON that matches
+//the expected JSON, however all arrays are treeted as un-ordered.
+//The JSONs are decoded and the resulting objects are compared via
+//reflect.DeepEqual so things like key-ordering and whitespace shouldn't matter.
+func MatchUnorderedJSON(json interface{}) types.GomegaMatcher {
+	return &matchers.MatchUnorderedJSONMatcher{
+		JSONToMatch: json,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


Benjamintf1:
Whoops, for some reason I removed matchXML rather then adding a matcher.

Benjamintf1:
Messed up on a few things. Worked locally, no clue why they worked locally, they shouldn't have.

onsi:
Hey @Benjamintf1 I think this would make for a great community matcher that I'd be happy to add to the [community matchers page](https://github.com/onsi/gomega/wiki) and/or the documentation.  I think it's a bit too specific for the core matcher library, however.  If you're up for creating and maintaining a repo for this I'd be happy to link to it.

Benjamintf1:
Cool, I'm gonna close this. I have a bit more work to do to make it feature complete/Polished, I'll add an issue when I complete it.

PR #222, opened 2017-06-23
fields: Remove unknown argument from example

apelisse:
It looks like the examples are out of date, and one of the argument has been removed.
Update the examples.


onsi:
Thanks!

PR #223, opened 2017-07-01
Remove else when not needed

spring1843:
Removes the extra else conditions when the block ends with a return and outdents the line.


onsi:
Thanks!

Issue #227, opened 2017-09-01
gexec: Start panics with non-pointer receiver implementation of io.Writer


Bo0mer:
# My setup
```
go version go1.9 darwin/amd64
```
Gomega version - `dcabb60a477c2b6f456df65037cb6708210fbb02`

# What I did
I run the program bellow (link to playground - https://play.golang.org/p/AF6vrN9IWs).

```go
package main

import (
	"os/exec"

	"github.com/onsi/gomega/gexec"
)

func main() {
	var d discard
	gexec.Start(exec.Command("whoami"), d, d)
}

type discard int

func (d discard) Write(p []byte) (int, error) {
	return len(p), nil
}
```

# What I expected to happen
The program exits cleanly.
# What actually happens
Panic.
```
panic: reflect: call of reflect.Value.IsNil on int Value

goroutine 1 [running]:
reflect.Value.IsNil(0x1101ba0, 0x129ee60, 0x82, 0x129ee60)
	/usr/local/go/src/reflect/value.go:995 +0x12e
github.com/onsi/gomega/gexec.Start(0xc4200bc160, 0x11d16a0, 0x129ee60, 0x11d16a0, 0x129ee60, 0x0, 0x0, 0x0)
	/Users/ivan/workspace/Go/src/github.com/onsi/gomega/gexec/session.go:81 +0x21a
main.main()
	/tmp/gexec/main.go:11 +0xef
exit status
```

The panic is caused by the following lines[0], since calling `IsNil` on non-pointers does not make sense.
```go
	if outWriter != nil && !reflect.ValueOf(outWriter).IsNil() {
		commandOut = io.MultiWriter(commandOut, outWriter)
	}

	if errWriter != nil && !reflect.ValueOf(errWriter).IsNil() {
		commandErr = io.MultiWriter(commandErr, errWriter)
	}
```

I can look into submitting a PR if you're ok with that.
Thanks,
Ivan

[0] https://github.com/onsi/gomega/blob/master/gexec/session.go#L81

Bo0mer:
Changing the code above to
```go
	if outWriter != nil {
		switch reflect.ValueOf(outWriter).Kind() {
		case reflect.Ptr:
			if !reflect.ValueOf(outWriter).IsNil() {
				commandOut = io.MultiWriter(commandOut, outWriter)
			}
		default:
			commandOut = io.MultiWriter(commandOut, outWriter)
		}
	}

	if errWriter != nil {
[...]

PR #228, opened 2017-09-01
Fix panic when starting session with non-pointer receiver writers

Bo0mer:
Closes https://github.com/onsi/gomega/issues/227.


williammartin:
This looks exactly the same as https://github.com/onsi/gomega/pull/235

Any idea why it was duplicated?

williammartin:
Ah, there is a test in #235 

Could we close this in favour of that PR @Bo0mer ?

PR #229, opened 2017-09-27
Add matcher to check shasum of a file

rizwanreza:
It would be handy if gomega supports the Shasum checking matcher. We have been using this in our codebase to compare the shasum of files.

### DIFF
--- a/matchers/match_sha_sum_matcher.go
+++ b/matchers/match_sha_sum_matcher.go
@@ -1,0 +1,56 @@
+package matchers
+
+import (
+	"crypto/sha1"
+	"fmt"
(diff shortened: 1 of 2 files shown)


williammartin:
Typically, we've tried to avoid adding quite specific matchers to gomega directly in favour of documenting community matchers here: https://github.com/onsi/gomega/wiki

It's just tricky to avoid the endless amount of matchers and potential maintenance that might come with them. If you'd like to maintain your matcher somewhere, I'm happy to update the list.

rizwanreza:
Thanks for the reply @williammartin

PR #235, opened 2017-11-02
Do no assume that writers are always pointers

### LINKED ISSUE #227: gexec: Start panics with non-pointer receiver implementation of io.Writer

jvshahid:
This can be the case if the user passes ioutil.Discard as the
writer. ioutil.Discard is just a primitive type (int) which will cause
Value.IsNil to panic

fixes #227

### DIFF
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"reflect"
 	"sync"
 	"syscall"
 
@@ -78,11 +77,11 @@ func Start(command *exec.Cmd, outWriter io.Writer, errWriter io.Writer) (*Sessio
 
 	commandOut, commandErr = session.Out, session.Err
 
-	if outWriter != nil && !reflect.ValueOf(outWriter).IsNil() {
+	if outWriter != nil {
 		commandOut = io.MultiWriter(commandOut, outWriter)
 	}
 
-	if errWriter != nil && !reflect.ValueOf(errWriter).IsNil() {
+	if errWriter != nil {
 		commandErr = io.MultiWriter(commandErr, errWriter)
 	}
 
--- a/gexec/session_test.go
+++ b/gexec/session_test.go
@@ -2,6 +2,7 @@ package gexec_test
 
 import (
 	"io"
+	"io/ioutil"
 	"os/exec"
 	"syscall"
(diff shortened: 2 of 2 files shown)


williammartin:
This looks exactly the same as https://github.com/onsi/gomega/pull/228

Is there a reason this was duplicated?

williammartin:
Ah there is an extra test in here, have requested to close the other in favour of this.

jvshahid:
yeah, and the fix is simpler than the one proposed in #228

PR #237, opened 2017-12-01
Add methods to terminate/kill sessions and wait

hoegaarden:
Added new methods for sessions to kill resp. terminate them and wait for
that to actually happen:
- KillAndWait
- TerminateAndWait

Main incentive for those is to make it a bit easier to implement parts
of the gexec.Session interface -- to fake certain aspects of it, without the
need of that fake to know about gexec.Session.

### DIFF
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -160,6 +160,19 @@ func (s *Session) Kill() *Session {
 }
 
 /*
+KillAndWait sends the running command a SIGKILL signal.  It waits for the process to exit.
+
+If the command has already exited, Kill returns silently.
+
+KillAndWait uses Kill and Wait under the hood and accepts the same timeout/polling intervals that Wait does.
+
+The session is returned to enable chaining.
+*/
+func (s *Session) KillAndWait(timeout ...interface{}) *Session {
+	return s.Kill().Wait(timeout...)
+}
+
+/*
 Interrupt sends the running command a SIGINT signal.  It does not wait for the process to exit.
 
 If the command has already exited, Interrupt returns silently.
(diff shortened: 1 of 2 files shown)


williammartin:
I'm not totally sure about this for a couple of reasons. 

Firstly, `XAndY` methods make me a little nervous because there's no end to the combinations, particularly as interfaces grow.

Secondly, this introduces a mismatch in return type for `Wait()` and these newly introduced methods. By wrapping up a call to `ExitCode()` inside a method that ostensibly just `Wait`ed, for me this violates the principle of least astonishment. It's probably not a huge deal because it's documented by the API but it does mean we'd preclude in future, being able to do:

```
session.TerminateAndWait().Buffer()
```

Also, if this PR was to be accepted, we should probably make the interface consistent and have:

```
func (s *Session) InterruptAndWait(timeout ...interface{}) int {
	return s.Interrupt().Wait(timeout...).ExitCode()
}
```

williammartin:
In fact, it's even slightly worse because functions already exist with those names:

https://github.com/onsi/gomega/blob/master/gexec/session.go#L222-L249

And they only `Wait`.

hoegaarden:
We will remove the `.ExitCode()` bit.
I just thought that would be nice, as a call to `TerminateAndWait()` would be a synchronous call to terminate the process and can therefore right away return the exit code. But, as we do not *need* that, I am happy to remove it.

hoegaarden:
@williammartin Could you give it another look?

williammartin:
Looks good to me, just would like `InterruptAndWait` and then happy to merge. Thanks!

hoegaarden:
We found a better way to test our thing and this implementation is not needed after all.

PR #241, opened 2017-12-02
Fix typos in matcher documentation

ivy:
I found a couple typos in the docs. Thanks for making Ginkgo and Gomega!

### DIFF
--- a/matchers.go
+++ b/matchers.go
@@ -269,7 +269,7 @@ func ContainElement(element interface{}) types.GomegaMatcher {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


williammartin:
Hi @ivy thanks for the commit. I actually think the `iff` statements might be correct as is: https://en.wikipedia.org/wiki/If_and_only_if but I've seen this confuse people before, and I don't think it _really_ helps that much here so I'm happy to merge.

ivy:
That makes a lot of sense. I recall seeing some other places that were more
obviously (to me at least) abbreviated. Would you accept a pull request to
expand them? Thanks for merging in the meantime.

PR #249, opened 2017-12-06
Fix `go vet` issues to prepare to Go 1.10

alex-slynko:
Go 1.9 runs `go vet` with `go test`


williammartin:
LGTM thanks for the linting work you are doing.

PR #253, opened 2017-12-11
Update panic message to help users debug

challiwill:
I ran into an issue where ghttp was on my gopath, but gomega was vendored. This resulted in my tests passing unexpectedly, and it was difficult for me to trace it back to the source of the problem.


williammartin:
LGTM thanks

PR #256, opened 2017-12-15
Allows multiple goroutines to call Build at the same time

jvshahid:


williammartin:
Not sure I have full context on this so correct me if I'm wrong but looks like this is a way to avoid clobbering the `GOPATH` env of `ginkgo` so multiples go routines can call `Build`. Looks fine if so but is there a way to test this?

jvshahid:
this change was earlier made to fix calling `Build` on go1.9, see https://github.com/onsi/gomega/pull/230 for more details. unfortunately i couldn't come up with a clean way to test this race condition. I feel a naive test would be too noisy and wouldn't add a lot of value

Issue #258, opened 2018-01-22
*gexec.Session panics when session.Command.Process is nil


hoegaarden:
The [`Kill()`](https://github.com/onsi/gomega/blob/003f63b7f4cff3fc95357005358af2de0f5fe152/gexec/session.go#L157) and [`Signal()`](https://github.com/onsi/gomega/blob/003f63b7f4cff3fc95357005358af2de0f5fe152/gexec/session.go#L194) methods call methods on `session.Command.Process` -- the `Process` can however be nil which leads to a panic.

In slack we discussed whether this panic is reasonable -- does it only even happen in situations in which you want the test to fail anyway? 

The answer is "no". Consider the following code:
```go
var (
  thing1 *gexec.Session
  thing2 *gexec.Session
)
AfterEach(func(){
  // Ensure all the things are stopped
 thing2.Terminate()
 thing1.Terminate()
})
It("has a sensible lifecycle", func(){
  thing1.Start()
  thing2.Start()

  // Point 1
  Expect(allMyThings.AreUp()).To(BeTrue()) 

  thing2.Terminate()
  thing1.Terminate()

  Expect(allMyThings.AreDown()).To(BeTrue())
})
```

The test tries to bring down both `thing`s. If the expectation at "Point 1" fails, because `thing2` could not be started (`thing1` however could!), we want `thing1` to be shut down before the test stops executing, hence the `AfterEach`.

If the expectation at "Point 1" succeeds, we want to go on to check that we can shut down both things cleanly, and leave the system in the state we expect.

If the `Terminate()` method returned an error instead of panicing, this test would be even clearer:

```go
var (
  thing1 *gexec.Session
  thing2 *gexec.Session
)
AfterEach(func(){
  // Ensure all the things are stopped. If they're already stopped ignore the errors
 thing2.Terminate()
 thing1.Terminate()
})
It("has a sensible lifecycle", func(){
  thing1.Start()
  thing2.Start()

  // Point 1
  Expect(allMyThings.AreUp()).To(BeTrue()) 

  Expect(thing2.Terminate()).To(Succeed())
  Expect(thing1.Terminate()).To(Succeed())

  Expect(allMyThings.AreDown()).To(BeTrue())
})
```

nodo:
Fixed in https://github.com/onsi/gomega/pull/259

PR #259, opened 2018-01-25
Avoid sending a signal or killing the session if the process is not alive

nodo:
Fix issue #258

### DIFF
--- a/gexec/session.go
+++ b/gexec/session.go
@@ -151,11 +151,7 @@ If the command has already exited, Kill returns silently.
 The session is returned to enable chaining.
 */
 func (s *Session) Kill() *Session {
-	if s.ExitCode() != -1 {
-		return s
-	}
-	s.Command.Process.Kill()
-	return s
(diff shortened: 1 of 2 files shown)


williammartin:
Looks sensible to me but I'm wondering if we can consolidate on using signal in Kill like:

```
func (s *Session) Kill() *Session {
	return s.Signal(syscall.SIGKILL)
}
```

Just seems we could get rid of an extra conditional at the cost of a slightly less lower abstraction, which I think on the whole is a win. WDYT?

nodo:
I think it makes sense @williammartin. 

I can also see that using `Signal` in `kill` is consistent with `Terminate` and `Interrupt` (https://github.com/onsi/gomega/blob/master/gexec/session.go#L161-L181). Even though they don't have a correspondent method in `process` (which is the case for `kill`), the code would like nicer.

williammartin:
LGTM thanks

PR #263, opened 2018-02-02
add unmarshalled matchers to gomega initial commit

Benjamintf1:
This is a barely modified version. I lowercased the package because it's super annoying. They still seem super verbose, I still need to add back the readme maybe from the original, and I'll change anything if you have stylistic preferences. I wanted to get this started before I went to sleep. 

Let me know what I need to do and I'll work on it quickly.


onsi:
heyo @Benjamintf1 I think you can just put this in with the existing matchers package and update the [DSL](https://github.com/onsi/gomega/blob/master/matchers.go).

Anyone object?

williammartin:
+1 To pulling this into the existing matchers package

Benjamintf1:
Some issues I wanted insight on before acting on(i see the failed tests, and I'll work on those too)

1. The deep match is pretty simular for matchjson, should I switch the default json matcher to use it or keep it with the old version for simplicity?

2. If order is not defined in some lists, it can be a real pain to test with json/yaml functions, however having all the functions is maybe a bit ridiculous, I'm thinking drop to 
MatchJson(, optional (with sets))
ContainJson(, optional with sets)
MatchYaml(, optional with sets)
ContainYaml(, optional with sets)

It leaves the names less cumbersome while still allowing all the features of the original package. 

3. Should yaml even be included, or should I extend to xml too?

4. Should I hide more of the matcher construction?

Benjamintf1:
I'm still working on the more ambitious and simpler version, but this should be the existing ones.

Benjamintf1:
@williammartin This is the  (i think) cleaner version of it. I removed the expanded yaml matcher because as I was working on it I realized the existing tests weren't accurate, didn't even test the new features, and I couldn't manage to get them to work(and I've seen utterly no demand for it). 

Any thoughts?

williammartin:
Will try to review this afternoon, thanks!

blgm:
Here are some thoughts of mine:
- passing the magic string `a` to change the behaviour of the matcher leads to tests that are not clear to read without reference to the documentation.  Perhaps it would be better to create a matcher with a different name, or take a named constant to change the behaviour?
 - there are some situations that the tests leave uncovered.  For instance, would ``Ω(`{"a":{"b":"c"}`).Should(ContainJSON(`{"b":"c"}`))`` pass or fail?  The behaviour is not defined in the existing tests.  I get that classical TDD does not necessarily lead to these tests being written, but well defined behaviours in edge cases are important for matchers.

Benjamintf1:
@blgm Thank you very much. 

1:  I was trying to use a "helper" function, ListsAsSets("a", "b"), but the list regularization would be undesirable. I could leave this feature out or in if it's more desirable, but I haven't heard either way. (perhaps even as an option as MatchJson().ListsAsSets() modifier?(either as a optional list, or just a switch but I haven't seen such a pattern before). Adding a separate matcher seems very heavy for this kind of feature(especially if combined with match and contain options). I've found that a lot of our json lists are actually sets(or without a defined order), which makes using matchjson difficult. Do you want me to remove this for now, and wait for a demand for it, or use one of those different approaches?

2: Absolutely, I can add a test for said behavior. The question you should be asking for a subset matching behavior is, I think, if a downstream consumer was trying to de-marshal your json payload and looking for a subset of the payload, would their code still function. It both allows de-coupling of small sections of your json-payload, as well as not breaking old tests as you add new features. If that additional test is added, would that be sufficient?

blgm:
1. I think it's a useful function to have, so I think it should be left in.  As long as the text `ListAsSets` or similar appears in the code, then people will be able to understand it.  I'm perfectly ok with a modifier.
1. I can see arguments either way for what the behaviour should be, and don't have a very strong opinion (perhaps in the future it could be modifiable anyway).  A test to pin down what the behaviour is today would be helpful - just so that no-one can break it in future without causing a test failure.

PR #265, opened 2018-02-12
Fix incorrect MatchElements examples

adamcohen:
This PR fixes the examples for `MatchElements` which were incorrect and didn't compile.


williammartin:
Awesome, thanks so much for fixing them up!

Issue #244, opened 2017-12-04
The Ω of Ω assertions in Ω tests


williammartin:
Ω is self perpetuating. Adding tests and adhering to conventions means using Ω. Let's wholesale, or suite-by-suite get rid of Ω assertions in favour of `Expect`.

It would be really great if we could fix this in the [docs](https://onsi.github.io/gomega/) too.

onsi:
Melancholy sigh
It is time to say goodbye
Farewell Ω

williammartin:
Looks like we can perhaps use something like:

```
for pattern in \
    'Ω(a).Should(b) -> Expect(a).To(b)' \
    'Ω(a).Should(b, c) -> Expect(a).To(b, c)' \
    'Ω(a).ShouldNot(b) -> Expect(a).NotTo(b)' \
    'Ω(a).ShouldNot(b, c) -> Expect(a).NotTo(b, c)' \
    'Ω(a).To(b) -> Expect(a).To(b)' \
    'Ω(a).ToNot(b) -> Expect(a).NotTo(b)' \
    'Ω(a).NotTo(b) -> Expect(a).NotTo(b)' \
; \
do find . -name '*.go' -a -not -path '*/Godeps/*' | xargs gofmt -w -r "${pattern}"; done
```

to achieve this...

james-lawrence:
just going to leave this here: `https://godoc.org/golang.org/x/tools/cmd/eg`

Benjamintf1:
https://github.com/onsi/gomega/pull/268

PR #268, opened 2018-02-15
change Ω to Expect in most places

Benjamintf1:
as per https://github.com/onsi/gomega/issues/244


Benjamintf1:
Left it in where it was compared to Expect in docs, and of course, the function definition itself.

kirederik:
Thanks @Benjamintf1 !

PR #272, opened 2018-03-01
fix typo floa32 > float32

trayo:


nodo:
Thanks for contributing @trayo! LGTM

PR #273, opened 2018-03-14
Make string pretty diff user configurable

### LINKED ISSUE #192: Make string equality diff comparison user-configurable
As of pr #190 the `Equal` matcher attempts to diff long strings. We should make this configurable so users can get more information when they need it.

One suggestion: put something in `format.go` like `MaxDepth` that allows users to change this behavior. This would be consistent with the current approach.

Another suggestion: use the builder pattern on the EqualMatcher to set some relevant configuration closer to the site of the matcher.

vayan:
Let the user set `format.TruncatedDiff = false` if he wants to display the full strings to get more informations about an equality failure.

Fix #192

Doc PR: https://github.com/onsi/gomega/pull/274

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -30,6 +30,9 @@ Set PrintContextObjects = true to enable printing of the context internals.
 */
 var PrintContextObjects = false
 
+// TruncatedDiff choose if we should display a truncated pretty diff or not
+var TruncatedDiff = true
+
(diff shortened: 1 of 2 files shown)


williammartin:
LGTM thanks, just commented a small thing on the doc PR.

scottschulthess:
awesome thank you, used this today!

PR #278, opened 2018-04-12
Use httputil.DumpRequest to pretty-print unhandled requests

benmoss:
Right now when the test server receives an unhandled HTTP request, it prints out default representation of the request object, which ends up being humongous (see https://gist.github.com/benmoss/86f69211817e33161aa678cb63e0ca57).

The httputil package offers a pretty printer for request objects that gives you the relevant details while sparing you the rest.

```
      Received Unhandled Request
      Expected
          <string>: POST /oauth/token HTTP/1.1
          Host: 127.0.0.1:51014
          Accept: application/json
          Accept-Encoding: gzip
          Content-Length: 97
          Content-Type: application/x-www-form-urlencoded
          User-Agent: Go-http-client/1.1

          client_id=test_client&client_secret=test_secret&grant_type=client_credentials&response_type=token
      to be nil
```

I can't see how we would encounter an error while dumping the request, but maybe someone more knowledgable than I might have an idea why this might be problematic.

### DIFF
--- a/ghttp/test_server.go
+++ b/ghttp/test_server.go
@@ -111,6 +111,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httputil"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


benmoss:
Looks like some flaky tests caused the build failure.

williammartin:
Hey I reran the tests, we're working on fixing flakiness atm. Thanks for the PR, seems to make sense, will have a closer look later.

williammartin:
LGTM thanks!

PR #279, opened 2018-04-20
Added first failure path matching for YAML

andrewgilbert12:
@abg and I noticed that currently the JSON matcher lists the path to the first failure when a match fails. This PR implements the same feature with the YAML matcher.

We noticed that @Benjamintf1 also implements something similar [here](https://github.com/Benjamintf1/Expanded-Unmarshalled-Matchers/blob/master/expanded_yaml_matcher.go).

### DIFF
--- a/matchers/match_json_matcher.go
+++ b/matchers/match_json_matcher.go
@@ -28,7 +28,7 @@ func (matcher *MatchJSONMatcher) Match(actual interface{}) (success bool, err er
 	json.Unmarshal([]byte(actualString), &aval)
 	json.Unmarshal([]byte(expectedString), &eval)
 	var equal bool
-	equal, matcher.firstFailurePath = deepEqual(aval, eval)
+	equal, matcher.firstFailurePath = deepEqualJSON(aval, eval)
 	return equal, nil
 }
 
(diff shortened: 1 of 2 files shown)


williammartin:
Hey @andrewgilbert12 sorry for taking so long to get round to looking at this. Is there a way to test this behaviour?

andrewgilbert12:
Hey @williammartin, I cribbed the JSONMatcher unit tests, seems like that's the only place where the deepEquals for JSON is being tested.

gcapizzi:
Hi @andrewgilbert12 ! This looks great. We've noticed a bit of duplication (e.g. `deepEqualJSON` and `deepEqualYAML` look very similar), would you like to give a shot at removing it? We're also happy to merge as-is and refactor later 🙂Let us know what you think!

/cc @alamages

andrewgilbert12:
I combined deepEqualJSON and deepEqualYAML into a single deepEqual function. It's currently just in the match_yaml_matcher.go file but I realize this should probably be in a separate file since it's also used in match_json_matcher.go. I'm not too familiar with the code structure so I didn't move it, but feel free to put the function somewhere that makes more sense.

gcapizzi:
Thanks @andrewgilbert12!

PR #281, opened 2018-05-10
AssignableToTypeOf returns false without error if exactly one of actual or expected is nil.

### LINKED ISSUE #280: BeAssignableToTypeOf fails with confusing error when just one of actual or expected is nil
The relevant lines of code are [here](https://github.com/onsi/gomega/blob/master/matchers/assignable_to_type_of_matcher.go#L15-L17).

Our code under test:
```go
func Configure() interface{} {
  return nil
}
```

Our test:
```go
It("returns a govmomi client", func() {
  expectedClient := new(govmomi.Client)

  actualClient := packageundertest.Configure()
  Expect(actualClient).To(BeAssignableToTypeOf(expectedClient))
})
```

Test output:
```
    returns a govmomi client [It]
    /Users/pivotal/go/src/github.com/rowanjacobs/terraform-provider-vsphereovf/vsphereovf/provider_test.go:68

    Refusing to compare <nil> to <nil>.
    Be explicit and use BeNil() instead.  This is to avoid mistakes where both sides of an assertion are erroneously
 uninitialized.
```

This confuses us since `expectedClient` is not actually nil (printing it and its address confirms this).

nmahoney-pivotal:
Fixes #280.

cc @rowanjacobs

### DIFF
--- a/matchers/assignable_to_type_of_matcher.go
+++ b/matchers/assignable_to_type_of_matcher.go
@@ -12,8 +12,10 @@ type AssignableToTypeOfMatcher struct {
 }
 
 func (matcher *AssignableToTypeOfMatcher) Match(actual interface{}) (success bool, err error) {
-	if actual == nil || matcher.Expected == nil {
+	if actual == nil && matcher.Expected == nil {
 		return false, fmt.Errorf("Refusing to compare <nil> to <nil>.\nBe explicit and use BeNil() instead.  This is to avoid mistakes where both sides of an assertion are erroneously uninitialized.")
+	} else if actual == nil || matcher.Expected == nil {
+		return false, nil
 	}
 
 	actualType := reflect.TypeOf(actual)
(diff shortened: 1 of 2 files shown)


blgm:
With this change:
```
Expect(nil).To(BeAssignableToTypeOf(nil)) // Error
Expect(17).To(BeAssignableToTypeOf(nil))  // OK
Expect(nil).To(BeAssignableToTypeOf(17))  // OK
```

I think it should be:
```
Expect(nil).To(BeAssignableToTypeOf(nil)) // Error
Expect(17).To(BeAssignableToTypeOf(nil))  // Error
Expect(nil).To(BeAssignableToTypeOf(17))  // OK
```
The matcher should be able to cope with whatever ends up in the `Expect()`.  For other situations you should either:
- check that it's assignable to a non-`nil` type, **OR**
- check that is `nil` using the `BeNil()` matcher

Any thoughts?

rowanjacobs:
Done!

PR #283, opened 2018-05-25
Fix data race in ghttp server

tylerschultz:
Accessing the HTTPTestServer variable can cause a data race.


luan:
@tylerschultz this looks good and it works. But I think I'd rather have an `RWMutex` and then `RLock()` in the two methods you added these since they just read from the variables. Or at the very least let's rename the mutex from `writeLock` since the name currently implies that there will be some writing.

What do you think?

tylerschultz:
Ya, considered it and took the minimalist approach. I'll change it.

luan:
Thanks @tylerschultz!

PR #286, opened 2018-06-05
allow 'Receive' matcher to be used with concrete types

jvshahid:
for example the following usage is now allowed:
```go
  var err *someError // *someError implement the error interface
  Expect(ch).To(Receive(&err)) // where ch is of type chan error
```


jvshahid:
turns out the purpose of the PR isn't very clear. This pr avoids the following pattern (from the [executor](https://github.com/cloudfoundry/executor/blob/23ff7f3487ee5b05110ca8839eb8b35d6a51f162/depot/steps/health_check_step_test.go#L118-L122)):

```go
var expectedError interface{}
Eventually(performErr).Should(Receive(&expectedError))
err, ok := expectedError.(*steps.EmittableError)
Expect(ok).To(BeTrue())
Expect(err.WrappedError()).To(Equal(disaster))

```

PR #291, opened 2018-06-19
Output concrete type when comparing slices of interface{}

nodo:
fixes #61


nodo:
Actually, I have realised this change doesn't make much sense. A slice of interface could be anything and it's not a good idea to show as concrete type the type of the first element. Possibly we could list some of the concrete types.... not sure anymore. 

I'll close the PR for now, let's discuss in the issue.

PR #293, opened 2018-07-20
show threshold in failure message of BeNumericallyMatcher

grosser:
solves https://github.com/onsi/ginkgo/issues/503#issuecomment-405161478
@williammartin

### DIFF
--- a/matchers/be_numerically_matcher.go
+++ b/matchers/be_numerically_matcher.go
@@ -13,11 +13,23 @@ type BeNumericallyMatcher struct {
 }
 
 func (matcher *BeNumericallyMatcher) FailureMessage(actual interface{}) (message string) {
-	return format.Message(actual, fmt.Sprintf("to be %s", matcher.Comparator), matcher.CompareTo[0])
+	return matcher.FormatFailureMessage(actual, false)
 }
 
 func (matcher *BeNumericallyMatcher) NegatedFailureMessage(actual interface{}) (message string) {
-	return format.Message(actual, fmt.Sprintf("not to be %s", matcher.Comparator), matcher.CompareTo[0])
+	return matcher.FormatFailureMessage(actual, true)
+}
+
+func (matcher *BeNumericallyMatcher) FormatFailureMessage(actual interface{}, negated bool) (message string) {
+	if len(matcher.CompareTo) == 1 {
+		message = fmt.Sprintf("to be %s", matcher.Comparator)
+	} else {
+		message = fmt.Sprintf("to be within %v of %s", matcher.CompareTo[1], matcher.Comparator)
+	}
+	if negated {
+		message = "not " + message
+	}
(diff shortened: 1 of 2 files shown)


matchers/be_numerically_matcher.go
```diff
@@ -13,11 +13,25 @@ type BeNumericallyMatcher struct {
...
+	var formatted string
+	if len(matcher.CompareTo) == 1 {
+		formatted = fmt.Sprintf("to be %s", matcher.Comparator)
+	} else {
+		formatted = fmt.Sprintf("to be within %v of %s", matcher.CompareTo[1], matcher.Comparator)
+	}
+	if !positive {
+		formatted = "not " + formatted
+	}
+	formatted = format.Message(actual, formatted, matcher.CompareTo[0])
```
williammartin: Can we just return here in place of setting the `formatted` variable and then returning?

matchers/be_numerically_matcher.go
```diff
@@ -13,11 +13,25 @@ type BeNumericallyMatcher struct {
...
+	return matcher.FormatFailureMessage(actual, true)
 }
 
 func (matcher *BeNumericallyMatcher) NegatedFailureMessage(actual interface{}) (message string) {
-	return format.Message(actual, fmt.Sprintf("not to be %s", matcher.Comparator), matcher.CompareTo[0])
+	return matcher.FormatFailureMessage(actual, false)
+}
+
+func (matcher *BeNumericallyMatcher) FormatFailureMessage(actual interface{}, positive bool) (message string) {
+	var formatted string
```
williammartin: Following on from the comment below, maybe we could just rename this to `message`, since it's the thing that _will_ be formatted, instead of the thing that _is_ formatted.

matchers/be_numerically_matcher.go
```diff
@@ -13,11 +13,25 @@ type BeNumericallyMatcher struct {
...
-	return format.Message(actual, fmt.Sprintf("to be %s", matcher.Comparator), matcher.CompareTo[0])
+	return matcher.FormatFailureMessage(actual, true)
 }
 
 func (matcher *BeNumericallyMatcher) NegatedFailureMessage(actual interface{}) (message string) {
-	return format.Message(actual, fmt.Sprintf("not to be %s", matcher.Comparator), matcher.CompareTo[0])
+	return matcher.FormatFailureMessage(actual, false)
+}
+
+func (matcher *BeNumericallyMatcher) FormatFailureMessage(actual interface{}, positive bool) (message string) {
```
williammartin: Could we maybe change `positive` here to `negated`, and flip the boolean expressions to be consistent with the Method names. Also, can we make this private `formatFailureMessage`?

grosser:
updated @williammartin

williammartin:
LGTM thanks.

PR #296, opened 2018-07-24
Add Module Definition

joefitzgerald:
This PR adds a `go.mod` file, turning the package into a go module. Go module support is introduced in `go 1.11` (currently in beta). Changes to `build_test.go` allow it to function correctly in the absence of an explicitly set `GOPATH`.


.travis.yml
```diff
@@ -5,10 +5,12 @@ go:
   - 1.8.x
   - 1.9.x
   - 1.10.x
+  - 1.11beta2
 
 install:
-  - go get -v ./...
+  - env GO111MODULE=on go get -v ./...
```
joefitzgerald: While this is redundant in go 1.11, it is left here so that the build works in prior versions of go.

terminatingcode: This is still needed in go 1.11 so a good idea all around to put in.

.travis.yml
```diff
@@ -5,10 +5,12 @@ go:
   - 1.8.x
   - 1.9.x
   - 1.10.x
+  - 1.11beta2
```
blgm: Now that go 1.11 is officially out, this can be updated to remove the 'beta'

gexec/build_test.go
```diff
@@ -60,28 +63,42 @@ var _ = Describe(".BuildWithEnvironment", func() {
...
 
 	BeforeEach(func() {
-		gopath = os.Getenv("GOPATH")
-		Expect(gopath).NotTo(BeEmpty())
-		Expect(os.Setenv("GOPATH", "/tmp")).To(Succeed())
-		Expect(os.Environ()).To(ContainElement("GOPATH=/tmp"))
+		original = os.Getenv("GOPATH")
+		gopath = filepath.Join(os.TempDir(), "gopath1")
+		target = "github.com/onsi/gomega/gexec/_fixture/firefly/"
+		Expect(os.MkdirAll(filepath.Join(gopath, "src", target), 0755)).To(Succeed())
```
blgm: Would it be possible to use temporary directories rather than hard-coded `gopath1`?  The reason is that there would be interference if two people on the same server were to run these tests at the same time.  I know it's unlikely, but it would be really hard to debug if it happened...

go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/onsi/gomega
+
+require (
+	github.com/fsnotify/fsnotify v1.4.7 // indirect
+	github.com/golang/protobuf v1.1.0
+	github.com/hpcloud/tail v1.0.0 // indirect
+	github.com/onsi/ginkgo v1.6.0
+	golang.org/x/net v0.0.0-20180724234803-3673e40ba225
```
blgm: We generated this file as part of doing a review, and got slightly different different timestamps/checksums.  Would you mind updating the file so that it reflects and official go version, rather than a beta?

blgm:
@joefitzgerald, we have cherry-picked the commit to master and added the small changes described above.  Thank you so much for your contribution to Gomega.
https://github.com/onsi/gomega/commit/a085d30924d5e2c520c4d00ea5c560f835fce1e9

go.mod
joefitzgerald: Google has not yet tagged versions for golang.org/x/net, unfortunately.

PR #298, opened 2018-08-25
Fix MatchJSON fail to parse json.RawMessage

### LINKED ISSUE #290: MatchJSON should support json.RawMessage
There is a type (`RawMessage`) in `json` package that is typed alias `[]byte`.
We use the type on a struct that is used for unmarshaling JSON.
When we tried to use `MatchJSON` that `RawMessage` field on a JSON string, we received the following error:

```
MatchJSONMatcher matcher requires a string, stringer, or []byte.  Got actual:
        	    <json.RawMessage | len:55, cap:64>: [{"clusters":[{"cluster":"pizza-boxes"}],"name":"AZ1"}]
```

It looks like the type casting of `actual` in `MatchJSON` tests for a `[]byte`, but not a typed aliased version of it.

hieven:
I add a type assertion for `json.RawMessage` because it's actually an alias of `[]byte`

fix #290

### DIFF
--- a/matchers/match_json_matcher_test.go
+++ b/matchers/match_json_matcher_test.go
@@ -1,6 +1,8 @@
 package matchers_test
 
 import (
+	"encoding/json"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/matchers"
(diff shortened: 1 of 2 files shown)


nodo:
Thanks @hieven,

Do you think you could add a test for that?

hieven:
@nodo 
sure, Just add a test for it.
Thanks for the feedback

nodo:
Thanks @hieven , LGTM!

PR #305, opened 2018-10-05
Fixed matcher for content-type

### LINKED ISSUE #304: ghttp.VerifyJSON doesn't work properly
While using `ghttp.VerifyJSON` I'm getting 
```
Expected
        <string>: application/json; charset=utf-8
    to equal
        <string>: application/json
```
`application/json; charset=utf-8` is a valid content-type for json and is used by default in some 3rd party http client libraries.

kucherenkovova:
fixes #304

### DIFF
--- a/ghttp/handlers.go
+++ b/ghttp/handlers.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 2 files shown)


williammartin:
LGTM thanks

Issue #295, opened 2018-07-20
failure messages are largely untested


grosser:
```
go-testcov github.com/onsi/gomega/matchers
ok  	github.com/onsi/gomega/matchers	0.324s	coverage: 85.7% of statements
matchers/have_key_matcher.go new uncovered sections introduced (6 current vs 0 configured)
matchers/have_key_matcher.go:38.84,39.28
matchers/have_key_matcher.go:40.20,41.69
matchers/have_key_matcher.go:42.10,43.60
matchers/have_key_matcher.go:47.91,48.28
matchers/have_key_matcher.go:49.20,50.73
matchers/have_key_matcher.go:51.10,52.64
.... 100 more lines or so
```

... I can open a few PRs to improve things where it's straight forward if there is any desire to get good code coverage ?

grosser:
did not look to closely, but definitely some of the failure messages

williammartin:
@grosser Did you have a think about submitting PRs for this?

grosser:
think about it: yes
have nothing else to work on: no :D

blgm:
@grosser, this issue has had little movement for about a year now. I'm in agreement with you that it's unfortunate that more failure messages are not tested. We are happy to review any more PRs that improve coverage. From a pragmatic perspective, I'm not sure there's much value in keeping this issue open as it's not driven much work in the past year, and having lots of inactive issues just causes noise for people trying to find and report bugs. I'm planning to close this issue off soon, but wanted to give you a chance to object if you'd like to.

grosser:
I wanted to come back to this but always found other things that are more pressing ... fine to close it 👍

PR #307, opened 2018-11-04
matcher coverage for BeZero and Succeed

grosser:
`go-testcov github.com/onsi/gomega/matchers` and then adding tests where needed

for https://github.com/onsi/gomega/issues/295

@williammartin looks good ? ... then I'll do another batch

### DIFF
--- a/matchers/be_zero_matcher_test.go
+++ b/matchers/be_zero_matcher_test.go
@@ -6,7 +6,7 @@ import (
 )
 
 var _ = Describe("BeZero", func() {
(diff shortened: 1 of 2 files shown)


williammartin:
This LGTM, I think it's good to capture this output under test. Thanks for doing the work on this. Also happy for you to do it in small pieces, since I guess there are a bunch of commands like this, if that works better for you.

grosser:
thx, next batch will be bigger, just wanted to make sure the direction is
ok before doing more :)

PR #316, opened 2018-11-23
more matcher tests

grosser:
for https://github.com/onsi/gomega/issues/295

@williammartin


matchers/match_xml_matcher.go
```diff
@@ -70,7 +70,7 @@ func parseXmlContent(content string) (*xmlNode, error) {
 			if err == io.EOF {
 				break
 			}
-			return nil, fmt.Errorf("failed to decode next token: %v", err)
+			return nil, fmt.Errorf("failed to decode next token: %v", err) // untested section
```
williammartin: Can you explain these commentas a little more? Is this saying there is no test coverage, but you don't want to test it because you are just doing error messages?

grosser:
there is no test coverage ... and I gave up on figuring out how to test
this since I got lost in the complexity of it :(

williammartin:
Cool cool. I think I'd prefer not to have comments in the code for this. If it's important enough to have a comment that is essentially a TODO, it's probably important enough to have an issue.

grosser:
These comments get read by go-testcov to ignore during coverage calculation,
I was thinking long-term it would be nice if tests failed when accidentally
uncovered code is added
(mark with the comment or add test-coverage to get tests green)
does that sounds useful to you?

williammartin:
Sorry for dropping the ball on this. Does the coverage tool you suggest report on the "ignored lines"? Could we turn on this coverage tool and then resolve the failures instead of marking them for later? Does the coverage tool report that these lines are "ignored"? Thanks.

grosser:
[go-testcov](https://github.com/grosser/go-testcov) does not report ignored lines, since I was thinking that that is not necessary since the ignored lines are already marked in the code / easy to find with `git grep`.

If we turn the coverage on now I need to add a bunch more ignores first to make it green.

grosser:
so in short: I'd prefer to merge this as it is since it's not too bad and then either continue on this piecemeal approach or do the "fail test on missing coverage with just making all uncovered" PR as you prefer

PR #326, opened 2019-01-21
mark all uncovered files so go-testcov ./... works

grosser:
WIP followup to https://github.com/onsi/gomega/pull/316

afaik options are:
 - replace Makefile `ginko` with `go-testcov ./...` (ugly formatting and all options gone)
 - add `go-testcov ./...` as an extra run (slows things down)
 - add `go-testcov ./...` as `make coverage`

atm a full run fails on ~25 files having coverage gaps, so added these `uncovered` declarations to make it pass (and fail if anything new uncovered is added)

@williammartin let me know how you like this or if you have ideas how to improve

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -1,6 +1,9 @@
 /*
 Gomega's format package pretty-prints objects.  It explores input objects recursively and generates formatted, indented output with type information.
 */
+
+// untested sections: 4
+
 package format
 
 import (
--- a/gbytes/say_matcher.go
+++ b/gbytes/say_matcher.go
@@ -1,3 +1,5 @@
+// untested sections: 1
+
 package gbytes
 
 import (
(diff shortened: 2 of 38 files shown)


williammartin:
I think my preference would be point 3. I don't want to put a dependency on `go-testcov` for everyone who wants to just run the tests.

williammartin:
I merged this because frankly, I'd like to be more optimistic about changes rather than just letting them stagnate (sorry about that!).

That said, I'm a bit concerned about these comment lines because it couples the source to some random coverage tool. I don't fully understand the semantics of what an "uncovered section" is, or how, as a contributor when I see a failure from `go-testcov` in Travis, I would know what my next steps are. I'll have a play around.

williammartin:
I also added https://github.com/onsi/gomega/commit/37ee382bd2be39d2a488b89e8f1f50c7d36ebf28 just to bring this up-to-date. If we want to continue with this idea we should probably get this into Travis asap so things don't get stale.

Issue #328, opened 2019-03-05
ginkgo assert equals does not escape special characters


pnikonowicz:
instead of displaying errors like this:
```
Expected
        <string>: some error

    to equal
        <string>: some error
```

display them like this:
```
  Expected
      <string>: "\x1b[31msome error\x1b[0m\n"
  to equal
      <string>: "some error"
```

williammartin:
Closed by #329

PR #329, opened 2019-03-05
display special chars on error

pnikonowicz:
previously, special chars would be rendered via the terminal
making it difficult to see errors

use %q to safely display special chars on the terminal
this addresses issue #328

Signed-off-by: Paul Nikonowicz <pnikonowicz@pivotal.io>


williammartin:
Makes sense to me, thanks!

PR #330, opened 2019-03-07
Xf/fix keys errors

xanderflood:
Small fix to my previous `MatchKeys` PR. That code had been forming the new failure messages using the `reflect.Value` version of the the `actual`, rather than the underlying interface, leading to obnoxious and uninformative error messages.

1. Fixed error messages in `keys.go`.
2. In the process I noticed that `reflect.Value.MapIndex` and `reflect.Value.Field` both guarantee to never return an invalid `reflect.Value`, so I removed what I believe to be a redundant check in `fields.go`. I'm happy to put it back if you disagree or prefer to include it for style reasons.

### DIFF
--- a/gstruct/fields.go
+++ b/gstruct/fields.go
@@ -118,12 +118,7 @@ func (m *FieldsMatcher) matchFields(actual interface{}) (errs []error) {
 				return nil
 			}
 
-			var field interface{}
-			if val.Field(i).IsValid() {
-				field = val.Field(i).Interface()
-			} else {
-				field = reflect.Zero(typ.Field(i).Type)
-			}
+			field := val.Field(i).Interface()
 
 			match, err := matcher.Match(field)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


williammartin:
Hey @xanderflood, thanks fro the new PR. Given that you've found some bad error messages from the matcher, is it possible to add a test to ensure they are now sensible?

xanderflood:
@williammartin great idea, I'll likely get on that later today. By the way, do you have any thoughts on what the issue is in `travis`? It looks like the failed tests are in the `matchers` package, which I didn't touch. Does that one depend on `gstruct` in some way that I disrupted? `make test` is passing for me locally right now

xanderflood:
Oh nevermind, it looks like the failing tests are all testing for correct error messages, so I'll take a closer look at those when I write my own

williammartin:
It looks like something has broken us between `1.12` and `gotip`

xanderflood:
I added tests to `gstruct`, and took a peak at the `gotip` failures. It looks like the failing tests are all related to string representations of `errors.errorString` structs. I can't find anything about what the `golang` change was that could have caused it, but it looks like those tests might need to be refactored to be a little more forgiving, potentially.

williammartin:
Thanks @xanderflood, I tracked down the Go change you can read about it in https://github.com/onsi/gomega/issues/332

Will review this later

Issue #332, opened 2019-03-07
Tests are failing on gotip


williammartin:
```
• Failure [0.002 seconds]
MatchErrorMatcher [It] shows negated failure message
/Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:111

  Expected
      <string>: Expected
          <*errors.errorString | 0xc0002ec750>: {
              s: "foo",
              frame: {
                  frames: [0x10a23b1, 0x164115e, 0x15e5f58],
              },
          }
      not to match error
          <string>: foo
  to contain substring
      <string>: {s: "foo"}
      not to match error
          <string>: foo

  /Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:115
------------------------------
• Failure [0.002 seconds]
MatchErrorMatcher [It] shows failure message
/Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:104

  Expected
      <string>: Expected
          <*errors.errorString | 0xc000324c00>: {
              s: "foo",
              frame: {
                  frames: [0x10a23b1, 0x1640bde, 0x15e5f58],
              },
          }
      to match error
          <string>: bar
  to contain substring
      <string>: {s: "foo"}
      to match error
          <string>: bar

  /Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:108
• Failure [0.002 seconds]
MatchErrorMatcher When asserting against an error [It] should succeed when matching with an error
/Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:21

  Expected
      <*errors.errorString | 0xc0001c7680>: {
          s: "an error",
          frame: {
              frames: [0x10a23b1, 0x163dd8c, 0x15c84fe],
          },
      }
  to match error
      <*errors.errorString | 0xc0001c76b0>: {
          s: "an error",
          frame: {
              frames: [0x10a23b1, 0x163df22, 0x15c84fe],
          },
      }

  /Users/pivotal/go/src/github.com/onsi/gomega/matchers/match_error_matcher_test.go:26
------------------------------
• Failure [0.002 seconds]
Equal When asserting equality between objects [It] should do the right thing
/Users/pivotal/go/src/github.com/onsi/gomega/matchers/equal_matcher_test.go:23

  Expected
      <*errors.errorString | 0xc0003fef90>: {
          s: "foo",
          frame: {
[...]

PR #333, opened 2019-03-07
Handle errors after introduction of frames

williammartin:
This is the best rudimentary solution I have for #332

### DIFF
--- a/matchers/match_error_matcher.go
+++ b/matchers/match_error_matcher.go
@@ -2,7 +2,6 @@ package matchers
 
 import (
 	"fmt"
-	"reflect"
 
 	"github.com/onsi/gomega/format"
 )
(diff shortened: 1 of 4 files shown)


williammartin:
It obviously broke everything not on tip. Some cleverer assertions required :)

matchers/match_error_matcher.go
```diff
@@ -22,8 +21,9 @@ func (matcher *MatchErrorMatcher) Match(actual interface{}) (success bool, err e
 
 	actualErr := actual.(error)
 
-	if isError(matcher.Expected) {
-		return reflect.DeepEqual(actualErr, matcher.Expected), nil
+	// deepEqual doesn't match errors anymore
+	if expectedErr, ok := matcher.Expected.(error); ok {
```
teddyking: I think in this block we can possibly also compare underlying types of the errors.

williammartin:
Looks like `reflect.DeepEqual` will be changed to ignore frames so we won't need any changes https://github.com/golang/go/issues/29934#issuecomment-472148465

PR #335, opened 2019-03-08
Add WithOffset methods to WithT type

markthemarkest:
This allows an offset to be specified for `WithT` methods.

There was not existing tests for the DSL so we didn't add any. ¯\\\_(ツ)\_/¯


markthemarkest:
We see that this PR is failing against the 'tip' release of Go. We noticed that in versions of Golang later than 1.9, we can use the 'Helper' function to achieve the desired results.

We're going to close this PR and approach this using the 'Helper' function instead.

williammartin:
I wouldn't worry about the `gotip` failure. I'm looking at this in https://github.com/onsi/gomega/issues/332. Nothing to do with your changes. But if there's already something that works for you, that's cool.

PR #343, opened 2019-04-17
Optimize contain_element_matcher

ahadas:
Create a function that returns the element at index i in
ACTUAL before iterating ACTUAL. The benefit of that is
threefold:
(A) It reduces the number of calls to 'isMap'
(B) It simplifies the code within the loop
(C) It narrows down the scope of the 'keys' variable


williammartin:
LGTM thanks

PR #344, opened 2019-04-17
Add BeElementOf matcher

ahadas:
This PR proposes the BeElementOf matcher that checks if actual is contained within the passed in
element(s). BeElementOf() always uses Equal() to perform the match.

When the passed in elements are comprised of a single element that is either a Map or Array or Slice, it behaves as the reverse form of the ContainElement() with the Equal() matcher. It is specifically useful in the following case:
```golang
Eventually(fn).Should(BeElementOf(<arr>))
```

Otherwise, it provides syntactic sugar for:
```golang
Or(Equals(1), Equals(2), ...)
```

### DIFF
--- a/matchers/be_element_of_matcher.go
+++ b/matchers/be_element_of_matcher.go
@@ -1,0 +1,55 @@
+package matchers
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/onsi/gomega/format"
+)
+
+type BeElementOfMatcher struct {
+	Elements []interface{}
+}
+
+func (matcher *BeElementOfMatcher) Match(actual interface{}) (success bool, err error) {
+	if reflect.TypeOf(actual) == nil {
+		return false, fmt.Errorf("BeElement matcher expects actual to be typed")
+	}
+
+	length := len(matcher.Elements)
+	valueAt := func(i int) interface{} {
+		return matcher.Elements[i]
+	}
+	// Special handling of a single element of type Array or Slice
+	if length == 1 && isArrayOrSlice(valueAt(0)) {
+		element := valueAt(0)
+		value := reflect.ValueOf(element)
+		length = value.Len()
+		valueAt = func(i int) interface{} {
+			return value.Index(i).Interface()
+		}
+	}
+
+	var lastError error
+	for i := 0; i < length; i++ {
+		matcher := &EqualMatcher{Expected: valueAt(i)}
+		success, err := matcher.Match(actual)
+		if err != nil {
+			lastError = err
+			continue
+		}
+		if success {
+			return true, nil
+		}
+	}
+
+	return false, lastError
+}
+
+func (matcher *BeElementOfMatcher) FailureMessage(actual interface{}) (message string) {
+	return format.Message(actual, "to be an element of", matcher.Elements)
+}
+
+func (matcher *BeElementOfMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	return format.Message(actual, "not to be an element of", matcher.Elements)
+}
--- a/matchers.go
+++ b/matchers.go
@@ -269,6 +269,22 @@ func ContainElement(element interface{}) types.GomegaMatcher {
 	}
 }
 
+//BeElementOf succeeds if actual is contained in the passed in elements.
+//BeElementOf() always uses Equal() to perform the match.
+//When the passed in elements are comprised of a single element that is either an Array or Slice, BeElementOf() behaves
+//as the reverse of ContainElement() that operates with Equal() to perform the match.
+//    Expect(2).Should(BeElementOf([]int{1, 2}))
+//    Expect(2).Should(BeElementOf([2]int{1, 2}))
+//Otherwise, BeElementOf() provides a syntactic sugar for Or(Equal(_), Equal(_), ...):
+//    Expect(2).Should(BeElementOf(1, 2))
+//
+//Actual must be typed.
+func BeElementOf(elements ...interface{}) types.GomegaMatcher {
+	return &matchers.BeElementOfMatcher{
+		Elements: elements,
+	}
+}
+
 //ConsistOf succeeds if actual contains precisely the elements passed into the matcher.  The ordering of the elements does not matter.
 //By default ConsistOf() uses Equal() to match the elements, however custom matchers can be passed in instead.  Here are some examples:
 //
--- a/matchers/be_element_of_matcher_test.go
+++ b/matchers/be_element_of_matcher_test.go
@@ -1,0 +1,56 @@
+package matchers_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/matchers"
+)
+
+var _ = Describe("BeElementOf", func() {
+	Context("when passed a supported type", func() {
+		It("should do the right thing", func() {
+			Expect(2).Should(BeElementOf([2]int{1, 2}))
+			Expect(3).ShouldNot(BeElementOf([2]int{1, 2}))
+
+			Expect(2).Should(BeElementOf([]int{1, 2}))
+			Expect(3).ShouldNot(BeElementOf([]int{1, 2}))
+
+			Expect(2).Should(BeElementOf(1, 2))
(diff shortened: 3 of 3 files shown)


ahadas:
A 'real-world' use of this matcher can be found in [KubeVirt](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2075)

ahadas:
@williammartin ping, could you please take a look at this one?

williammartin:
Hey @ahadas, sorry I've been totally overwhelmed recently. Thanks for the PR! I have a couple of questions/thoughts as I look at this.

Firstly, in trying to understand the real-world example, are you using the `Eventually` purely as a timeout (i.e. the `func` blocks and there's no retry)? It doesn't look like the `func` can return a "null" value and I don't think `phases` is updated anywhere else so I guess this is the case. Just want to make sure I have a clear picture of the intent.

Secondly, do you have a _need_ for this to work across maps? I think you've decided this should be a counterpart to `ContainsElement` which seems consistent, but I also don't _love_ that it works for maps either. At least for `ContainsElement`, you can consider an array as a map of index => value, but with `BeIn` I wonder whether anyone might expect it to match a `key` (compared to say, `BeElementIn`). No strong feelings, just curious about your thoughts.

Thirdly, do you have a real-world example of the non-collection matching behaviour? Given an example from your tests:

```
Expect("abc").Should(BeIn(MatchRegexp("t"), MatchRegexp("a.*c")))
```

What makes this different than say:

```
Expect("abc").To(SatisfyAny(MatchRegexp("t"), MatchRegexp("a.*c")))
```

Thanks again!

ahadas:
> Hey @ahadas, sorry I've been totally overwhelmed recently. Thanks for the PR! I have a couple of questions/thoughts as I look at this.
> 

Sure, no problem :)

> Firstly, in trying to understand the real-world example, are you using the `Eventually` purely as a timeout (i.e. the `func` blocks and there's no retry)? It doesn't look like the `func` can return a "null" value and I don't think `phases` is updated anywhere else so I guess this is the case. Just want to make sure I have a clear picture of the intent.
> 

Well, every second, `func` is called and its return value is checked against `phases`. This happens for `seconds` seconds. So yeah, `phases` does not change but there is a "retry" mechanism that checks the returned phase couple of times to see if it matches one of the phases until reaching the timeout. 
 
> Secondly, do you have a _need_ for this to work across maps? I think you've decided this should be a counterpart to `ContainsElement` which seems consistent, but I also don't _love_ that it works for maps either. At least for `ContainsElement`, you can consider an array as a map of index => value, but with `BeIn` I wonder whether anyone might expect it to match a `key` (compared to say, `BeElementIn`). No strong feelings, just curious about your thoughts.
> 

Good point. No, I don't have a need for this to work across maps and I see your point. It may well not be clear whether actual is compared against the keys or against the values (If you look at my branch name _be_element_of_ - that was the reason for that :) in the beginning, I tried to make this matcher a complete mirror of `ContainElement` but then I thought about providing a syntactic sugar for ```Expect(X).To(Or(Equal(A), Equal(B),...)``` with ```Expect(X).To(BeIn(A, B)``` and that changed my view).

So do we want to raise an error when the matcher operates on a map?

> Thirdly, do you have a real-world example of the non-collection matching behaviour? Given an example from your tests:
> 
> ```
> Expect("abc").Should(BeIn(MatchRegexp("t"), MatchRegexp("a.*c")))
> ```
> 
> What makes this different than say:
> 
> ```
> Expect("abc").To(SatisfyAny(MatchRegexp("t"), MatchRegexp("a.*c")))
> ```
> 

No, I don't. Thinking out loud:
On the one hand, I think that supporting a non-collection matching is nice for the case I mentioned above (with `Equal` matchers) and that is easily implemented within the `BeIn` matcher.
On the other hand, I agree that `SatisfyAny` better fits the case of specifying the matchers explicitly.
In the beginning, I limited `BeIn` to the `Equal` matcher (in the first commit) and generalized it to other matchers later (in the second commit). 
The reason I did this generalization was to fix an [ugly hack I did](https://github.com/kubevirt/kubevirt/pull/2203/commits/d28a196057f16673e828ce0264504cee70f4995f) (Don't tell the KubeVirt guys :) ) - I wanted to check whether a string is matched by one of unknown number of regexps so I put the matching logic inside a function that compared its boolean return value. What I wanted to achieve is something like:
```golang
Expect(DefaultMachineType).Should(BeIn(X))
```
Where `X` is a collection of regexps that are generated from the comma-separated string `DefaultEmulatedMachines`.

How about generalizing `SatisfyAny` (and probably `SatisfyAll` as well) to get Arrays and Slices to support this collection matching and change `BeIn` to work only with `Equal`?

ahadas:
> So do we want to raise an error when the matcher operates on a map?
> 

Please look at the last changes, maps are now treated as any other type that is not an Array or a Slice (i.e., compared using `EqualMatcher`) - I think it makes the most sense but let me know what you think (btw, all commits can be squashed with the first one).

> How about generalizing `SatisfyAny` (and probably `SatisfyAll` as well) to get Arrays and Slices to support this collection matching and change `BeIn` to work only with `Equal`?

If that is the right way to go, I'll do it in a separate PR.

ahadas:
@williammartin Hi, could you please follow up on this one?

matchers/be_in_matcher_test.go
```diff
@@ -0,0 +1,56 @@
...
+			Expect(2).Should(BeIn([2]int{1, 2}))
+			Expect(3).ShouldNot(BeIn([2]int{1, 2}))
+
+			Expect(2).Should(BeIn([]int{1, 2}))
+			Expect(3).ShouldNot(BeIn([]int{1, 2}))
+
+			Expect(2).Should(BeIn(1, 2))
+			Expect(3).ShouldNot(BeIn(1, 2))
+
+			Expect("abc").Should(BeIn("abc"))
```
williammartin: This is a slightly confusing test and I wonder whether it also points to this really meaning `BeElementIn`. I could imagine trying to write:

`Expect("abc").To(BeIn("123abc456"))`

williammartin:
Hi @ahadas! Just to set some expectations (I should have done this before, sorry!), I maintain this in my own time, so the level to which I can respond fluctuates wildly. I was on vacation all last week. I'm totally fine with you pinging me as reminders, it's useful to me, just want to set expectations correctly.

I think that the matcher looks good now, with the exception of my comment on the naming. I think if the need arises later we can extend with more functionality but let's keep it scoped down for now. 

What do you think about the naming? Happy to move forward with this once we reach consensus on that.

matchers/be_in_matcher_test.go
ahadas: Yeah, I see your point, that might be confusing indeed. Changed.

ahadas:
> Hi @ahadas! Just to set some expectations (I should have done this before, sorry!), I maintain this in my own time, so the level to which I can respond fluctuates wildly. I was on vacation all last week. I'm totally fine with you pinging me as reminders, it's useful to me, just want to set expectations correctly.
> 

No worry about that, pinged just so you'll notice it's ready for a follow-up :)

> I think that the matcher looks good now, with the exception of my comment on the naming. I think if the need arises later we can extend with more functionality but let's keep it scoped down for now.
> 
> What do you think about the naming? Happy to move forward with this once we reach consensus on that.

Well, as I can't think of a better name let's switch it back to BeElementOf then

matchers/be_element_of_matcher.go
```diff
@@ -0,0 +1,55 @@
...
+		if success {
+			return true, nil
+		}
+	}
+
+	return false, lastError
+}
+
+func (matcher *BeElementOfMatcher) FailureMessage(actual interface{}) (message string) {
+	return format.Message(actual, "to be in", matcher.Elements)
```
williammartin: Should we be changing this to "to be an element of"? Similarly for the negated failure message?

ahadas: Yes, of course, missed that - Done

williammartin:
LGTM thanks!

williammartin:
Would you be able to write some docs for this on the `gh-pages` branch?

ahadas:
> Would you be able to write some docs for this on the `gh-pages` branch?

Sure, will do

PR #357, opened 2019-07-19
Remove unnecessary conversions

muesli:
sample_01 is already a byte slice. Those conversions don't make any difference.

### DIFF
--- a/matchers/match_xml_matcher_test.go
+++ b/matchers/match_xml_matcher_test.go
@@ -43,9 +43,7 @@ var _ = Describe("MatchXMLMatcher", func() {
 		})
 
 		It("should work with byte arrays", func() {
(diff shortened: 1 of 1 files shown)


williammartin:
I agree with you, but it seems like we should just remove this test. The assertion is already made in other tests.

blgm:
@muesli, thank you for spotting this problem. Because we haven't heard from you in a while I'm going to merge this PR so that you get GitHub contributor credit, and then immediately delete the test as suggested by @williammartin above to remove duplication. Thank you for you helping to make Gomega better!

muesli:
@blgm Thank you, that's very kind! Also I'm sorry, I didn't realize you were waiting for my action or feedback. What @williammartin mentioned sounded absolutely reasonable, I was just expecting other to chime in or a clear signal how to progress with this PR. Thanks for resolving it!

PR #362, opened 2019-10-17
Bump go-yaml version to cover fixed ddos heuristic

petrkotas:
This PR bumbs go-yaml to v2.2.4, which has the ddos vulnerability fixed.

**Issue:** 
go-yaml preceding 2.2.4 had vulnerability to ddos attack via billion laughs bomb.
Such attack lead to program to be unresponsive.
Issue has been described in https://raesene.github.io/blog/2019/10/15/From-stackoverflow-to-CVE/


blgm:
Hi @petrkotas. Thank you for submitting this. I notice that `go.mod` now has the line `go 1.13`. We want people on all supported versions of Go to be able to use Gomega, and I'm concerned that this might cause problems for people using older supported versions of Go. Would you mind removing that line? I realise that it was added automatically.

petrkotas:
Hi @blgm of course, sorry for that. Honestly I did not realised it is there since vscode keep adding this to my go.mod everywhere.

blgm:
Thank you @petrkotas!

Issue #351, opened 2019-05-24
Lazy evaluate the `optionalDescription` parameter


neowulf:
Thanks for the great library! We have been extensively using this library for our tests.

Would it be possible to not eagerly evaluate `errors.Cause(err).Error()` as part of the `optionalDescription` in the following statement:

	Expect(err).To(Not(HaveOccurred()), errors.Cause(err).Error())

This statement causes a panic when `err` is nil. But, when there's an error, it'd be useful to know the cause / stacktrace. The following would be a workaround but be nice to avoid it.

	errString := ""
	if err != nil {
		errString= errors.Cause(err).Error()
	}
	Expect(err).To(Not(HaveOccurred()), errString)

blgm:
It's not possible for Gomega to change how Go works, and I think Go will evaluate all the arguments to `To()` before calling it (please do correct me if I'm wrong). However in #193 it has been suggested that the `optionalDescription` could be a function, which could be lazily evaluated.

Issue #193, opened 2017-01-21
Allow optionalDescription to be a function in Expect().To()


tamasd:
It would be nice if I could call `Expect().To()` with a `func() string` as an optionalDescription parameter. This function would be evaluated only if the assertion fails.

My exact use case is that I do an HTTP request and assert the response code first, and if that fails, I would like to dump the body to stdout for easier debugging.

Currently I do it with a boolean variable and a defer, but it would be great to have a cleaner solution.

williammartin:
Sorry it took so long to get a response. This seems sensible and not sure it's too tricky. The relevant code is here: https://github.com/onsi/gomega/blob/master/internal/assertion/assertion.go#L46-L53 and it already takes `interface{}`. I guess maybe we'd only want to [build the description](https://github.com/onsi/gomega/blob/master/internal/assertion/assertion.go#L57) on failure in the case of a function.

Happy to review a PR for this, thanks.

PR #364, opened 2019-11-16
Allow optional description to be lazily evaluated function

ansd:
As suggested in #193 and #351, allow a single argument of type `func() string` to be passed as an optional description. This function gets lazily evaluated if there's a failure and the returned string is used to annotate the failure message.

### DIFF
--- a/internal/assertion/assertion.go
+++ b/internal/assertion/assertion.go
@@ -49,19 +49,20 @@ func (assertion *Assertion) NotTo(matcher types.GomegaMatcher, optionalDescripti
 }
 
 func (assertion *Assertion) buildDescription(optionalDescription ...interface{}) string {
-	switch len(optionalDescription) {
-	case 0:
+	if len(optionalDescription) == 0 {
 		return ""
-	default:
-		return fmt.Sprintf(optionalDescription[0].(string), optionalDescription[1:]...) + "\n"
 	}
+	if describe, ok := optionalDescription[0].(func() string); ok && len(optionalDescription) == 1 {
+		return describe() + "\n"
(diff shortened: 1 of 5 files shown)


internal/assertion/assertion.go
```diff
@@ -49,19 +49,20 @@ func (assertion *Assertion) NotTo(matcher types.GomegaMatcher, optionalDescripti
...
 
 func (assertion *Assertion) buildDescription(optionalDescription ...interface{}) string {
-	switch len(optionalDescription) {
-	case 0:
+	if len(optionalDescription) == 0 {
 		return ""
-	default:
-		return fmt.Sprintf(optionalDescription[0].(string), optionalDescription[1:]...) + "\n"
 	}
+	if describe, ok := optionalDescription[0].(func() string); ok && len(optionalDescription) == 1 {
```
jim-slattery-rs: It might benchmark as faster to do the original `switch len(optionalDescription)`, with this new `if` condition in the `case 1:` block.
That way you check the length once, and only attempt the type assertion `.(func() string)` when length is known to be 1.

jim-slattery-rs:
This looks like a great addition! Very handy.

internal/assertion/assertion.go
ansd: thanks for reviewing @jim-slattery-rs. It makes sense what you suggest, we pushed another commit with the change :)

blgm:
Looks great to me too.  Thanks @ansd!

nkuacac:
It seems the lazily evaluated function lost.
image
the "descs" not updated.

PR #365, opened 2019-11-16
Allow optional description to be lazily evaluated function

ansd:
Docs for #364


blgm:
Thanks @ansd. I appreciate that you keep the docs up to date! Too often I forget.

PR #370, opened 2019-12-08
Add ContainElements matcher

### LINKED ISSUE #361: ContainsElements matcher
I was surprised not to find an existing matcher called `ContainsElements` that asserts that a set of elements are a subset of another, similar to `ConsistOf`.  Does such a matcher exist?  Since I couldn't find it, I defined the matcher below for my tests, but it would only work with slices and arrays and not maps:

```golang
// ContainElements returns whether a slice contains all the elements in another slice
func ContainElements(elements interface{}) types.GomegaMatcher {
	value := reflect.ValueOf(elements)
	ms := make([]types.GomegaMatcher, 0, value.Len())
	for i := 0; i < value.Len(); i++ {
		ms = append(ms, &matchers.ContainElementMatcher{
			Element: value.Index(i).Interface(),
		})
	}
	return &matchers.AndMatcher{
		Matchers: ms,
	}
}
```

Would you be open to adding a `ContainsElements` matcher functionally similar to the above but the more complete error reporting and support for maps as well?  Thanks.

ansd:
Resolves #361.

This PR is based on top of the changes done in #369.

### DIFF
--- a/matchers/consist_of.go
+++ b/matchers/consist_of.go
@@ -16,11 +16,6 @@ type ConsistOfMatcher struct {
 	extraElements   []interface{}
 }
 
-var neighbours = func(v, m interface{}) (bool, error) {
-	match, err := m.(omegaMatcher).Match(v)
-	return match && err == nil, nil
-}
-
 func (matcher *ConsistOfMatcher) Match(actual interface{}) (success bool, err error) {
 	if !isArrayOrSlice(actual) && !isMap(actual) {
 		return false, fmt.Errorf("ConsistOf matcher expects an array/slice/map.  Got:\n%s", format.Object(actual, 1))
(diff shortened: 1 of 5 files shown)


matchers/consist_of.go
```diff
@@ -16,35 +16,18 @@ type ConsistOfMatcher struct {
 	extraElements   []interface{}
 }
 
+var neighbours = func(v, m interface{}) (bool, error) {
```
blgm: Maybe this should be defined as a function rather than a variable? This would make it constant so that you couldn't have bugs where it gets assigned to a different value.

ansd: thanks :)

matchers/contain_elements_matcher_test.go
```diff
@@ -0,0 +1,85 @@
...
+			It("should soldier on", func() {
+				Expect([]string{"foo", "bar", "baz"}).ShouldNot(ContainElements(BeFalse(), "foo", "bar"))
+				Expect([]interface{}{"foo", "bar", false}).Should(ContainElements(BeFalse(), ContainSubstring("foo"), "bar"))
+			})
+		})
+	})
+
+	Context("when passed exactly one argument, and that argument is a slice", func() {
+		It("should match against the elements of that argument", func() {
+			Expect([]string{"foo", "bar", "baz"}).Should(ContainElements([]string{"foo", "baz"}))
```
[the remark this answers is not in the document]
ansd: thanks @nodo, I added the negative test to both

PR #372, opened 2019-12-11
Change JSONRepresenting matcher to check MIME type

williammartin:
This brings `VerifyJSONRepresenting` inline with `VerifyJSON`.

Can add tests if we decide this is the correct way to go.


williammartin:
In a classic case of `git blaming` myself, I merged the change to `VerifyJSON` without updating `VerifyJSONRepresenting` a year ago: https://github.com/onsi/gomega/issues/304

PR #373, opened 2019-12-17
gliveexec exec with realtime output handling 

distorhead:
Package provides execution of external commands with realtime output handling and ability to fail fast when expectation of output was not met. Example of usage:

```
It("Should ...", func(done Done) {
    gotDeletingHookLine := false

    Expect(gliveexec.ExecCommand(dir, "mycmd", gliveexec.ExecCommandOptions{
        OutputLineHandler: func(line string) {
            Expect(strings.HasPrefix(line, "│ NOTICE Will not delete Job/migrate: resource does not belong to the helm release")).ShouldNot(BeTrue(), fmt.Sprintf("Got unexpected output line: %v", line))

            if strings.HasPrefix(line, "│ Deleting resource Job/migrate from release") {
                gotDeletingHookLine = true
            }
        },
    })).Should(Succeed())

    Expect(gotDeletingHookLine).To(BeTrue())

    close(done)
}, 300)
```

### DIFF
--- a/gliveexec/gliveexec.go
+++ b/gliveexec/gliveexec.go
@@ -1,0 +1,119 @@
+package gliveexec
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/onsi/gomega/gexec"
+)
+
+// ExecCommandOptions is an options for ExecCommand
+type ExecCommandOptions struct {
+	Env               map[string]string
+	OutputLineHandler func(string)
+}
+
+// ExecCommand allows handling output of executed command in realtime by CommandOptions.OutputLineHandler.
+// User could set expectations on the output lines in the CommandOptions.OutputLineHandler to fail fast
+// and give immediate feedback of failed assertion during command execution.
+func ExecCommand(dir, binPath string, opts ExecCommandOptions, arg ...string) error {
+	cmd := exec.Command(binPath, arg...)
+	cmd.Dir = dir
+
+	cmd.Env = os.Environ()
+	for k, v := range opts.Env {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	stdoutReadPipe, stdoutWritePipe, err := os.Pipe()
+	if err != nil {
+		return fmt.Errorf("unable to create os pipe for stdout: %s", err)
+	}
+
+	stderrReadPipe, stderrWritePipe, err := os.Pipe()
+	if err != nil {
+		return fmt.Errorf("unable to create os pipe for stderr: %s", err)
+	}
+
+	outputReader := io.MultiReader(stdoutReadPipe, stderrReadPipe)
+
+	session, err := gexec.Start(cmd, stdoutWritePipe, stderrWritePipe)
+	if err != nil {
+		return fmt.Errorf("error starting command: %s", err)
+	}
+
+	go func() {
+		<-session.Exited
+
+		// Initiate EOF for consumeOutputUntilEOF
+		stdoutWritePipe.Close()
+		stderrWritePipe.Close()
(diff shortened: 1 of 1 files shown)


williammartin:
Hi @distorhead,

Thank you for the PR. I'd like to understand this a little bit more. It looks like you are providing a new package that will Execute an external process, and then allow you to provide a function that will be provided with each line of output for you to make assertions on. Is that correct? Is there anything I'm missing?

distorhead:
> Hi @distorhead,
> 
> Thank you for the PR. I'd like to understand this a little bit more. It looks like you are providing a new package that will Execute an external process, and then allow you to provide a function that will be provided with each line of output for you to make assertions on. Is that correct? Is there anything I'm missing?

Hi!

Yes that's correct, handle/check output of the process line-by-line in realtime while the process is working, like unix pipelines a bit.

It is of course controversial decision to make a separate package in this particular way. But I made this PR mainly to present an idea of such realtime executor, which we found useful, when testing some long-running app with rich output.

williammartin:
Sure, we definitely appreciate coming with PRs to present ideas, thank you.

This may be slightly tangential, but also may highlight a misunderstanding on my part. I don't totally understand the example given above:

```
Expect(strings.HasPrefix(line, "│ NOTICE Will not delete Job/migrate: resource does not belong to the helm release")).ShouldNot(BeTrue(), fmt.Sprintf("Got unexpected output line: %v", line))

            if strings.HasPrefix(line, "│ Deleting resource Job/migrate from release") {
                gotDeletingHookLine = true
            }

...

    Expect(gotDeletingHookLine).To(BeTrue())
```

I can't see how this test would ever pass, as both prefixes cannot evaluate to `true`.

distorhead:
This check ensures that line with specified text will not be received at all during command execution (and test will fail fast if this line has been detected):

```
            Expect(strings.HasPrefix(line, "│ NOTICE Will not delete Job/migrate: resource does not belong to the helm release")).ShouldNot(BeTrue(), fmt.Sprintf("Got unexpected output line: %v", line))
```

And this check ensures that some line of the output will match specified text:

```
                if strings.HasPrefix(line, "│ Deleting resource Job/migrate from release") {
                    gotDeletingHookLine = true
                }
...
            Expect(gotDeletingHookLine).To(BeTrue())
```

The first check is impossible to do with standard gexec without waiting for external command termination.The second check is possible to do with standard gexec, it was given just for example completeness.

williammartin:
I wonder if you could use `buffer.Detect()` to capture the first case?

```
session := gexec.Start("cmd")
buffer := session.Buffer()
select {
case <-buffer.Detect("│ NOTICE Will not delete Job/migrate: resource does not belong to the helm release"):
	Fail("Got unexpected output line")
}
```

But there are probably 3 differences here:
 - I guess you'd need to introduce another channel to select on for when the process exits
 - You might not be able to get useful output (like the line you were printing before).
 - You can't use the Prefix matcher (though `Detect` does take a regex)

For context, I'm just poking at the problem here from different angles. I see why this is valuable, but I'm hesitant to introduce a whole new package and construct if we might be able to enhance the existing packages to support it.

blgm:
This has sat around for quite a while so I'm going to close it. Here's why:

- although this provides useful functionality, it's quite a niche requirement
- if we provide more than one way of doing the same thing, it takes more effort for new users to learn Gomega
- there is already a family of packages that work with Gomega, and I feel that this is a better fit as something that people can optionally use. If we find that it's getting very heavy usage then we can merge it into Gomega at a later stage. This saves the effort of maintaining it if there turn out to be few users.

I would happily add it to this list: https://github.com/onsi/gomega/wiki
However it doesn't seem to have an independent repo at the moment, so I can't do that yet

blgm:
Thanks for offering this contribution @distorhead. If you publish it as a standalone package then let me know and I will link to it.

PR #388, opened 2020-05-20
Doc of NegatedFailureMessage had a typo

tmaflexera:
If the `GomegaMatcher` was testing the `ShouldNot` case, `NegatedFailureMessage` will be called if `Match` returns `true`, not `false` as currently documented.


blgm:
Thank you @tmaflexera!

PR #399, opened 2020-10-28
Adding code to run on Powersystem

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: go
+arch:
+    - amd64
+    - ppc64le
 
 go:
   - 1.14.x


williammartin:
Hi @genisysram thanks for your contribution.

I don't think we make any guarantees of supporting gomega on power pc right now, but it's something we could look into. I see that you're working on behalf of IBM, are they looking for support?

genisysram:
I'm working on behalf of IBM, they will support power system.

williammartin:
@genisysram we can merge this so that IBM has confidence that gomega currently works on PowerPC but I suspect it's unlikely that any of the current maintainers will do anything if the tests break unless we see real usage.

genisysram:
we will look in future how it goes ,if code breaks , but merge as of now , such that we can say that the code will work on powersystems

PR #400, opened 2020-11-18
Add dockerfile and docker-compose for build environment

YuviGold:
Now a contributor can run `make test` without requiring a `go` installed or `ginkgo` package.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM golang:1.15
+
+RUN go get github.com/onsi/ginkgo/ginkgo
+RUN go install github.com/onsi/ginkgo/ginkgo
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
+
+.PHONY: test
 test:
 	[ -z "`gofmt -s -w -l -e .`" ]
 	go vet
 	ginkgo -p -r --randomizeAllSpecs --failOnPending --randomizeSuites --race
 
-.PHONY: test
+docker_test:
+	docker-compose build test && docker-compose run --rm test make test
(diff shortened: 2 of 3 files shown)


blgm:
Hi @YuviGold. You've identified two issues:
1. contributors need `ginkgo` to be installed
2. contributors need `go` to be installed

Regarding (1), an alternative approach would be to use `go run github.com/onsi/ginkgo/ginkgo`, and this would make sure that the version used is the same one as in `go.mod`

Regarding (2), it feels like a dependency on `go` is being replaced with a dependency on `docker-compose`. While I'm sure there are plenty of people who might think that's an improvement, there may be an equal number of people who think it's a bad idea. Given that Gomega is a Go project, I think most contributors would expect Go to be a pre-requisite, but might be more surprised if `docker-compose` was a pre-requisite.

I like this way of using `docker-compose` to manage the test environment, but I think it should be an optional way that some people choose to develop, rather than the default way that people are expected to develop.

YuviGold:
Hey @blgm.
Sometimes I like to contribute to technologies that I don't necessarily have their engines on my computer (rust, golang, ts, etc.)
So at least for me, having a `Dockerfile` with a `docker-compose.yaml` is the best option just to dive in and start running stuff.

But I can understand that not everyone likes this extra dependency on their computer (not to mention other container runtime engines that won't work, which requires specifically `docker` being installed).

I guess it could be added as another approach alongside the existing one i.e. `docker_test` Makefile target.
WDYT?

blgm:
I'm much happier with that. The "canonical" way to run tests is with `make test` and docker-compose is an option for those who want it. We could add other options in the future. If we do add other options it would be nice to find a way to move the configuration files into subdirectories, but I think it's ok to leave these ones in the main directory for now.

PR #407, opened 2021-01-12
Remove dependency on golang.org/x/xerrors

tklauser:
The `errors.Is` func and support for error wrapping using the `%w` verb are
available since Go 1.13, see https://blog.golang.org/go1.13-errors

The `go.mod` file already specifies Go 1.14 as the minimum supported
version, so it's safe to drop the transitional `golang.org/x/xerrors`
dependency.


tklauser:
The Travis CI failure on linux/ppc64le/Go 1.15.x seems unrelated (net/http: TLS handshake timeout): https://travis-ci.org/github/onsi/gomega/jobs/754087736

onsi:
thanks @tklauser !

PR #408, opened 2021-01-31
fix: collections matchers should display type of expectation

blgm:
The collections matchers ConsistOf() and ContainElements() are often
used with collections of consistent type, for example:

Expect(t).To(ConsistOf(1, 2, 3))

When this fails, it's more helpful to report:

   ...to contain elements <[]int | len:3, cap:3>: [1, 2, 3]

Than the previous behavior:

   ...to contain elements <[]interface {} | len:3, cap:3>: [1, 2, 3]

By using []interface{} only where there's no other option we make it
easier to debug issues - especially for types that display identically.

### DIFF
--- a/matchers/consist_of.go
+++ b/matchers/consist_of.go
@@ -81,6 +81,29 @@ func matchers(expectedElems []interface{}) (matchers []interface{}) {
 	return
 }
 
+func presentable(elems []interface{}) interface{} {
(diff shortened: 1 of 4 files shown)


onsi:
hey @blgm - this looks good to me.  ready for me to merge it in (or you can too, of course!)

PR #409, opened 2021-02-02
fix: print value type for interface{} containers

### LINKED ISSUE #161: Comparing interface{}
When comparing two items for equivalence I get this error:

```
  Expected
      <[]interface {} | len:2, cap:2>: [1, 2]
  to be equivalent to
      <[]interface {} | len:2, cap:2>: [1, 2]
```

I only get this when deserializing a []int{1, 2} from messagepack. I'm sure the types are at fault here, but the message could improve IMHO.

### LINKED ISSUE #385: Display more information for "BeEquivalentTo"
When I run a test with:
```golang
Expect(XXXX).Should(BeEquivalentTo(expected))
```
I had the following output with a test failed
```
Expected
          <map[string]interface {} | len:2>: {"super": "hero", "maxHP": 500000}
      to be equivalent to
          <map[string]interface {} | len:2>: {"super": "hero", "maxHP": 500000}
```

This is true, but I had a hard time figured out what was the real problem.
My property `maxHP` has been an `int` for expected and a `float64` by variable `XXXX`.

I would be nice, if an additional message could be provided to point out the mismatch in type.

Anyway, I like your product!

blgm:
When a value is contained in an interface{} we should show the type of
the value as well as the interface{} type.

Consider the test:

  Expect([]interface{}{1}).To(Equal([]interface{}{uint(1)}))

Previous message:

  Expected
    <[]interface {} | len:1, cap:1>: [1]
  to equal
    <[]interface {} | len:1, cap:1>: [1]

New message:

  Expected
    <[]interface {} | len:1, cap:1>: [<int>1]
  to equal
    <[]interface {} | len:1, cap:1>: [<uint>1]

By showing the type, it's much easier to work out what is going on.

Fixes #161 
Fixes #385

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -271,7 +271,7 @@ func formatValue(value reflect.Value, indentation uint) string {
 		}
 		return formatStruct(value, indentation)
 	case reflect.Interface:
-		return formatValue(value.Elem(), indentation)
+		return formatInterface(value, indentation)
 	default:
 		if value.CanInterface() {
 			return fmt.Sprintf("%#v", value.Interface())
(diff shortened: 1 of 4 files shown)


onsi:
this looks great - i think it's going to help alleviate a lot of confusion!

PR #411, opened 2021-02-06
feat(gexec) Add CompileTest functions. Close #410

### LINKED ISSUE #410: gexec could build test packages
Hello folks,

### Context

I would like to run [*conformance tests* from opencontainers](https://github.com/opencontainers/distribution-spec/tree/master/conformance) in the suite test of [harbor-operator](https://github.com/goharbor/harbor-operator).

Following the conformance documentation, the recommended way is to compile the test using `go test -c`.

### Expected behavior

I tried to use the `gexec` package to compile tests.
I found the `BuildXXX` functions to compile go packages, but nothing to compile tests.

I suggest to add new functions called `BuildTest(packagePath string, args ...string) (compiledPath string, err error)`

WDYT?

holyhope:
Adds followinf methods:
- CompileTest(packagePath string, args ...string) (compiledPath string, err error)
- CompileTestWithEnvironment(packagePath string, env []string, args ...string) (compiledPath string, err error)
- CompileTestIn(gopath string, packagePath string, args ...string) (compiledPath string, err error)

Signed-off-by: Pierre Péronnet <pierre.peronnet@ovhcloud.com>

### DIFF
--- a/gexec/_fixture/firefly/main_test.go
+++ b/gexec/_fixture/firefly/main_test.go
@@ -1,0 +1,7 @@
+package main_test
+
+import "testing"
+
+func Test(t *testing.T) {
+	t.Log("Hum, it seems okay.")
+}
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -3,6 +3,8 @@
 package gexec
 
 import (
+	"crypto/md5"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"go/build"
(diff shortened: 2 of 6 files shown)


holyhope:
I changed name of functions of #410. I suggested `BuildTest...`, but implemented `CompileTest...` so it matches the `-c` flags of `go test` and there is no confusion between a `TestBuild(*testing.T)` method and the one I implemented here.

holyhope:
Thanks to the last commit, `gexec` is now able to get test package before compile it :-)

holyhope:
ping @onsi

onsi:
whoops sorry!  thanks for the ping (and the code!) - will take a look today.

blgm:
I've had a very cursory look at this and did have two thoughts (which are very much thoughts rather than objections):
- with regard to `GetAndCompileTest()`, is there an argument that `Get()` could be independent, so it could be chained with `CompileTest()` or with something else?
- `$GOPATH` is on the way to being deprecated. I can see that this PR builds consistency with previous Gexec behaviour, and I do like being consistent. But I was wondering whether we might end up deprecating some of these functions quite soon when (or if) `$GOPATH` is removed?

holyhope:
Hello @blgm,

Thank you for taking care of my PR 👍

- `Get()` Could be independent but atm it will be used only for `CompileTest()`.
  An valid argument would be to implement more go commands such as `go vet`?
  Not sure about the terms *chained with*, I did not imagine to have chained method `Get().CompileTest()`, I rather expected
  ```golang
  func Test() {
    Get(packageName)
    CompileTest(packageName)
  }
  ```
  I agree that the first one looks nice. :-)
  I am not sure that a new structure should be introduced to handle the result of `Get()`.
  WDYT, should I introduce this structure? Suggestion:
  ```golang
  type GoGet interface{
    CompileTest(args ...string) (string, error)
    CompileTestWithEnv(envs []string, args ...string) (string, error)
    Build(args ...string) (string, error)
    BuildWithEnv(envs []string, args ...string) (string, error)
  }
  
  func Get(packagePath) GoGet
  ```
  **BTW in order to nicely chain methods, the `GoGet` objct will need to store the error in order to be raised with a final functions.**
  
- About `$GOPATH` I have no idea, I am sorry to tell you that I did not event know that it is deprecated. ¯\\\_(ツ)\_/¯

onsi:
Thanks @holyhope - I expect the $GOPATH stuff will need some attention at some point.  For now I'm happy to pull this in as is.

PR #419, opened 2021-02-24
feature: add index to gstruct element func

zach-source:
This is a feature I've been missing and worked around with a custom counter struct. The use case is I have a small slice and just want to be able to reference the objects by index. I either sort the list or use matchers that don't care about order. I was able to add a way to make the index available while keeping backwards compatibility.

## Changes

Add interface Identify - allow us to create interface funcs on top of existing code for backwards compatibility and flexibility.

Add new function IdentifierWithIndex. This identifier accepts an index along with element so you can use it in your string return value.

Add two new functions MatchElementsWithIndex, MatchAllElementsWithIndex. Essentially the same function as the original but swaps in the new IdenitiferWithIndex function so you can use the index.

Helper function IndexIdentify to provide an already created func for easy index use.

## Final words

Happy to make any changes necessary. Just thought it would be helpful. Thanks!

### DIFF
--- a/gstruct/elements.go
+++ b/gstruct/elements.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime/debug"
+	"strconv"
 
 	"github.com/onsi/gomega/format"
 	errorsutil "github.com/onsi/gomega/gstruct/errors"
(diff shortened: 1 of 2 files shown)


onsi:
This looks good to me and makes sense.  Do you mind adding some documentation on the gh-pages branch to [this section](https://onsi.github.io/gomega/#testing-type-slice) as a separate PR?

THanks!

zach-source:
@onsi Awesome, don't mind at all.

zach-source:
https://github.com/onsi/gomega/pull/421 opened for documentation.

blgm:
@onsi I see that you merged the documentation PR #421, did you mean to merge this implementation PR too, or was there something outstanding?

onsi:
lol omg - thanks for the ping @blgm !!

zach-source:
@onsi @blgm was also thinking the same thing but I just assumed you were waiting for a reason. Thanks for merging my contribution!

PR #422, opened 2021-02-25
Allow WithTransform()'s function to accept a nil value

jim-slattery-rs:
Add a unit test to verify this use case.

### DIFF
--- a/matchers/with_transform.go
+++ b/matchers/with_transform.go
@@ -40,15 +40,25 @@ func NewWithTransformMatcher(transform interface{}, matcher types.GomegaMatcher)
 }
 
 func (m *WithTransformMatcher) Match(actual interface{}) (bool, error) {
-	// return error if actual's type is incompatible with Transform function's argument type
-	actualType := reflect.TypeOf(actual)
-	if !actualType.AssignableTo(m.transformArgType) {
-		return false, fmt.Errorf("Transform function expects '%s' but we have '%s'", m.transformArgType, actualType)
+	// prepare a parameter to pass to the Transform function
+	var param reflect.Value
+	{
+		if actual != nil {
+			// return error if actual's type is incompatible with Transform function's argument type
+			actualType := reflect.TypeOf(actual)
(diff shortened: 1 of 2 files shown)


onsi:
awesome, thanks!

matchers/with_transform.go
```diff
@@ -40,15 +40,25 @@ func NewWithTransformMatcher(transform interface{}, matcher types.GomegaMatcher)
...
+			// return error if actual's type is incompatible with Transform function's argument type
+			actualType := reflect.TypeOf(actual)
+			if !actualType.AssignableTo(m.transformArgType) {
+				return false, fmt.Errorf("Transform function expects '%s' but we have '%s'", m.transformArgType, actualType)
+			}
+			param = reflect.ValueOf(actual)
+		} else {
+			// make a nil value of the expected type
+			param = reflect.New(m.transformArgType).Elem()
+		}
```
cbandy, line 56: This is no longer checking that the `actual = nil` is appropriate for the transform function. This incorrectly succeeds:

```go
gomega.WithTransform(
	func(int) bool { return true },
	gomega.BeTrue(),
).Match(nil)
```

I recommend the following:

```go
	var param reflect.Value
	if actual == nil && m.transformArgType.Kind() == reflect.Interface {
		// The original type of actual is unknown, so there's no way to make its
		// reflect.Value. Create a nil of the transform argument, which is known.
		param = reflect.Zero(m.transformArgType)

	} else if actual != nil && reflect.TypeOf(actual).AssignableTo(m.transformArgType) {
		// The original type of actual is compatible with the transform argument.
		param = reflect.ValueOf(actual)

	} else {
		return false, fmt.Errorf("Transform function expects '%s' but we have '%T'", m.transformArgType, actual)
	}
```

onsi: thanks @cbandy - are you up for submitting this as a PR?

cbandy: Yep, no problem. Submitted #423.

PR #423, opened 2021-02-28
Check more carefully for nils in WithTransform

### LINKED ISSUE #420: WithTransform of nil error panics
The following panics in v1.10.5:

```go
var err error
gomega.WithTransform(
	func(error) bool { return true },
	gomega.BeTrue(),
).Match(err)
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x56b58b]

goroutine 1 [running]:
github.com/onsi/gomega/matchers.(*WithTransformMatcher).Match(0xc00005e380, 0x0, 0x0, 0x7d5a18, 0xc00005e380, 0xc00006a058)
        /home/cbandy/go/pkg/mod/github.com/onsi/gomega@v1.10.5/matchers/with_transform.go:45 +0x6b
main.main()
```

I encountered it while using `Eventually()` to watch for a specific error, e.g.

```go
Eventually(func () error { ... }).Should(WithTransform(os.IsNotExist, BeTrue()))
```

cbandy:
Follow-up to #422 which fixes #420.

### DIFF
--- a/matchers/with_transform.go
+++ b/matchers/with_transform.go
@@ -42,18 +42,17 @@ func NewWithTransformMatcher(transform interface{}, matcher types.GomegaMatcher)
 func (m *WithTransformMatcher) Match(actual interface{}) (bool, error) {
 	// prepare a parameter to pass to the Transform function
 	var param reflect.Value
-	{
-		if actual != nil {
-			// return error if actual's type is incompatible with Transform function's argument type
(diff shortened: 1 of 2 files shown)


onsi:
thanks!

PR #432, opened 2021-03-31
Fix error message in BeNumericallyMatcher

khernyo:
When passing a non-number as the threshold, the error message should
contain the problematic threshold, not the expected value.


khernyo:
The failures seem to be unrelated.

PR #437, opened 2021-04-25
Add Satisfy() matcher

### LINKED ISSUE #425: How to match error with boolean function?
I want to test whether my `err` is "not exist", I found [`os.IsNotExist()`](https://pkg.go.dev/os#IsNotExist) in the standard library. How can I use it in my tests? I'm wondering something like
```go
Expect(err).To(MatchError(os.IsNotExist))
```
But that doesn't work.

kaoet:
Satisfy() takes a predicate function and matches the actual value
against it.

```go
isEven := func (i int) bool { return i%2==0 }
Expect(2).To(Satisfy(isEven))
```

Fixes #425


onsi:
Hey thanks for this @kaoet - looks great!

PR #441, opened 2021-04-30
Bump golang.org/x/net

skitt:
This doesn’t change anything significant for Gomega, but it reduces
the number of versions in the dependency tree, and helps get rid of
CVE flags in projects using Gomega and Ginkgo (in particular,
CVE-2020-9283 on golang.org/x/crypto, and CVE-2020-14040 on
golang.org/x/text).

Signed-off-by: Stephen Kitt <skitt@redhat.com>


blgm:
Thank you @skitt!

PR #446, opened 2021-05-25
gexec.Build should create a binary with an expected name

aminjam:
Changes introduced in onsi/gomega#410 is breaking backward compatibility for building binaries with an expected name.

### DIFF
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -3,8 +3,6 @@
 package gexec
 
 import (
-	"crypto/md5"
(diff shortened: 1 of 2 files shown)


blgm:
This makes sense to me. I have seen it break backwards compatibility, although it's not strictly a breaking change since there was never a guarantee of the binary name. If there was a clear benefit from having a pseudo-random binary name, then it might be worth it, but I can't really see any advantage at all.

PR #450, opened 2021-06-08
Fix example of `MatchAllElementsWithIndex`

autopp:
The `MatchAllElementsWithIndex`'s example should use `MatchAllElementsWithIndex`, but uses `MatchAllElements`.
The current example contains type errors.


onsi:
Thank you!

PR #454, opened 2021-07-07
Call Verify GHTTPWithGomega receiver funcs

### LINKED ISSUE #453: ghttp Internal Server Error using `NewWithT(t)` and `VerifyJSON`
If you set up a test using the ghttp server with `NewWithT` and `NewGHTTPWithGomega`, then use the `VerifyJSON` handler, the ghttp server responds with 500. (For more background, see https://github.com/onsi/gomega/issues/321#issuecomment-447950666 for an example of using ghttp with the XUnit style of tests calling `NewWithT`.)

Here's an example with `github.com/onsi/gomega@v1.13.0`:

```golang
package main

import (
	"bytes"
	. "github.com/onsi/gomega"
	"github.com/onsi/gomega/ghttp"
	"mime"
	"net/http"
	"testing"
)

func Test(t *testing.T) {
	g := NewWithT(t)
	gh := ghttp.NewGHTTPWithGomega(g)

	server := ghttp.NewServer()
	defer server.Close()

	server.AppendHandlers(
		ghttp.CombineHandlers(
			gh.VerifyRequest("POST", "/"),
			gh.VerifyJSON("{}"),
			gh.RespondWith(http.StatusCreated, "OK"),
		),
	)

	r, _ := http.Post(server.URL(), mime.TypeByExtension(".json"), bytes.NewBuffer([]byte("{}")))
	g.Expect(r.StatusCode).To(Equal(http.StatusCreated))
}
```

Running the test:
```
 $ go test
--- FAIL: Test (0.00s)
    main_test.go:28: 
        Expected
            <int>: 500
        to equal
            <int>: 201
FAIL
exit status 1
```

I expect the test to pass instead of fail. 

What I think is happening is that the `VerifyMimeType` call in `VerifyJSON` is using the package level `gomega` instead of using the `GHTTPWithGomega` struct that's passed around.

https://github.com/onsi/gomega/blob/9da0d1334f01b8e0868ac7b7e663c350d45f4661/ghttp/handlers.go#L134

I think this call should be `g.VerifyMimeType("application/json")` to use the correct `gomega`. 

I'm happy to submit a PR to fix this so that the verify receiver functions call other receiver functions. Let me know! :smiley:

macaptain:
The GHTTPWithGomega struct has receiver functions to allow XUnit style
tests to use the ghttp server. However, some of these receiver functions
call out to Verify functions that use the package level gomega instead
of the struct's gomega.

Fixes #453

### DIFF
--- a/ghttp/handlers.go
+++ b/ghttp/handlers.go
@@ -109,7 +109,7 @@ func (g GHTTPWithGomega) VerifyHeader(header http.Header) http.HandlerFunc {
 //(recall that a `http.Header` is a mapping from string (key) to []string (values))
 //It is a convenience wrapper around `VerifyHeader` that allows you to avoid having to create an `http.Header` object.
 func (g GHTTPWithGomega) VerifyHeaderKV(key string, values ...string) http.HandlerFunc {
-	return VerifyHeader(http.Header{key: values})
+	return g.VerifyHeader(http.Header{key: values})
 }
 
 //VerifyBody returns a handler that verifies that the body of the request matches the passed in byte array.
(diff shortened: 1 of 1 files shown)


onsi:
Thanks @macaptain !

PR #455, opened 2021-07-07
add caching for go modules

jtarchie:
When Github Actions run, it is downloading the go modules each run. It looks like some failures are happening because of rate limiting on some go modules.

This will use the `actions/cache` Github Action to cache the golang modules.


onsi:
thanks!  here's hoping it helps!

jtarchie:
Did it?

Issue #471, opened 2021-09-27
allow WithTransform transform functions to report transformation errors instead of panic'ing


TheDiveO:
As I'm using `WithTransform` more and more lately to write domain-specific matchers I tend to hide ugly low-level details in them, such as working with both a type `T` as well as `*T`, or even hiding the ugly face of Go's embedding (which is not a working substitute for inheritance). From looking at examples as well as The Source I see that `panic`ing seems to be the current mode. This looks rather heavy-handed to me and not in the spirit of failing matchers, especially as it terminates all tests.

Would you find transform functions being able to report back errors a worthwhile feature to include with Gomega? This looks as if this can be added in a fully backwards-compatible manner: `WithTransform` would accept a transformer function that either:

1. takes exactly one parameter and returns exactly one value (as before),
2. additionally takes exactly one parameter and returns exactly two values, the second being of type error.

A transform function encountering an error when trying to transform its actual value would then follow the Go Way of Errors and not panic but instead return an `error`-type value.

onsi:
Thanks @TheDiveO I agree that being able to return an error would be a better approach here.  Thanks for the PR.  I've merged it in.

PR #472, opened 2021-09-27
allow transform functions to report errors

TheDiveO:
- additionally accepts second error-type transform function return value
- adds test cases
- updates WithTransform godoc

Okay, so I already managed to ignore the first highlighted bullet point in CONTRIBUTING.md 🤦🏼 So, here's issue #471 .

I ran `make test` to do the formatting checks, vetting, and finally unit testing with success.


onsi:
This is fantastic - thank you!

Issue #392, opened 2020-09-17
RFC on error assertions for functions with multi-return values and trailing error value


TheDiveO:
This is a request for comment on extending Gomega to more smoothly handle (phrase) error assertions for functions which return multiple values, including a trailing error. While working on my Gomega 3rd party extension "errxpect" (https://github.com/TheDiveO/errxpect) I noticed some things and I would like to get feedback on ways to do it better or if gomega could be opened to 3rd party implementors of `gomega.Assertion`.

So, instead of:

```go
    _, _, err := foo(42)
    Expect(err).To(MatchError("DOH!"))
```

simply spell it out as:

```go
    Errxpect(foo(42)).To(MatchError("DOH!"))
```

Things I noticed and had to work around:

- Gomega's internal implementation of `gomega.Assertion` basically does two separate things of relevance to our discussion here:
  - it implements the important logic for calling a fail handler and generating useful stack context information -- which is something that any 3rd party implementation of the `gomega.Assertion` probably would like to tap into ("inherit").
  - specific multi-return value checking in form of a single actual parameter, and a extra... parameters. Please note that this check conflicts with my specific usecase of having a non-zero error, or a zero error with potentially non-zero other return values.

- Gomega's `ExpectWithOffset` pattern cannot be applied to multi-return value functions. This actually can be easily remedied with a generic `WithOffset(o int)` function added to the `gomega.Assertion` interface. In my errexpect module I return a more specially typed assertion object, which then offers `WithOffset(...)`, so I don't need to touch Gomega's `Assertion` interface.

My questions for feedback:

- At the moment I have to get Gomega's assertion functionality for triggering fail handlers and producing stack context information by running a transient, separate assertion when checking the multiple return values for calling the user's matcher(s). Could you imagine opening up the functionality for triggering fail handlers and stack context traces to 3rd party implementers of `Assertion`?

- With me coming up with `Errxpect` I'm throwing stones in my glasshouse ... but since `ExpectWithOffset(...)` slightly stutters and with Option receivers being all the rage, could you imagine a generic `Assertion.WithOffset(o)`, so Gomega's assertions/expectations can also be spelled out as `Expect(...).WithOffset(1).To(...)`?

onsi:
Hey @TheDiveO sorry for the delay.

[...]

PR #478, opened 2021-11-01
separate out offsets and timeouts

TheDiveO:
This is a (late) follow-up to #392 and separates control over offset, timeout and polling intervals from the multitude of expectation-related function signatures. In consequence, test writers can stick solely to the baseline of `Expect`, `Eventually`, `Consistently` in the future:
* `WithOffset(offset int)` on (async) assertions sets the offset, so `ExpectWithOffset(666, ...).To(FooBar())` can now be written as `Expect(...).WithOffset(666).To(FooBar())`. The same applies to `Eventually` and `Consistently`.
* async assertions additionally feature `WithTimeout(interval time.Duration)` and `WithPolling(interval time.Duration)`. Here, I opted for a type-safe interface and not repeating "Interval" in the method names, as the function parameter names already carry this meaning.
* existing interfaces are kept unmodified.
* `internal/` tests are updated to use `WithOffset`, `WithTimeout` and `WithPolling`.
* godoc comments have been updated.

### DIFF
--- a/gomega_dsl.go
+++ b/gomega_dsl.go
@@ -204,7 +204,8 @@ func Expect(actual interface{}, extra ...interface{}) Assertion {
 //    ExpectWithOffset(1, "foo").To(Equal("foo"))
 //
 // Unlike `Expect` and `Ω`, `ExpectWithOffset` takes an additional integer argument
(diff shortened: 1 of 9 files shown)


onsi:
This looks great - thanks so much!

PR #479, opened 2021-11-01
fix transformation error reporting

TheDiveO:
* fix transformation error reporting using an invalid %e formatting direction.
* updates corresponding unit test to catch such formatting directive errors.
* corrects documentation to correctly refer to `WithTransformation`.


onsi:
Good catch - thanks!

PR #480, opened 2021-11-03
add Error() assertions on the final error value of multi-return value functions

TheDiveO:
This PR follows up on #392, implementing a new `Error` method on `Assertions`.
- adds `Error` method to `Assertion`, returning a new `Assertion` configured for error-related assertions (see below for details).
- refactors Assertion to store both the actual value as well as all extras in a single slice for uniform handling.
- refactors extra values vetting into generic slice of actual values vetting with optionally skipping a particular element, either the first element for normal assertions and the last element for error assertions.
- an `Error`-returned `Assertion` is configured to vet differently from non-error-related assertions in order to correctly handle the Go error idiom, where either (A) all return values must be zero values in case of a trailing non-nil error, or (B) there are no constraints on the return values when the trailing error value is nil.
- changes vet-related error messages to refer to "Unexpected non-nil/non-zero argument ..." instead of "Unexpected non-nil/non-zero **extra** argument".
- adapts existing tests to the new vet-related error messages.
- adds new test table covering `Error`.

### DIFF
--- a/internal/assertion.go
+++ b/internal/assertion.go
@@ -8,17 +8,22 @@ import (
 )
 
 type Assertion struct {
-	actualInput interface{}
+	actuals     []interface{} // actual value plus all extra values
+	actualIndex int           // value to pass to the matcher
+	vet         vetinari      // the vet to call before calling Gomega matcher
 	offset      int
-	extra       []interface{}
 	g           *Gomega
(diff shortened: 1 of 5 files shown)


onsi:
Looks great, thanks!

PR #482, opened 2021-11-05
document WithTransform error reporting from #471

TheDiveO:


onsi:
sweet - thanks!  I'll cut a release with all this stuff soon :)

Issue #484, opened 2021-11-18
[RFC] new Value matcher?


TheDiveO:
I would like to get feedback before venturing into this area: in several projects (not least [@thediveo/lxkns](https://github.com/thediveo/lxkns)) I want to express assertions regarding high-level objects in a also comparably highly cross-references information model.

The custom domain-specific matchers should be versatile to use and especially when used in different contexts, test spec writers shouldn't need to worry whether a `ContainElement(...)` will need a `PointTo` or not. Especially when combining lower abstraction matchers into higher abstraction matchers, using and not using `PointTo` becomes a distracting low-level and cumbersome aspect. This is even more painful as the Go compiler itself hides direct and indirect accesses in many places, so we only need to write `x.y` without caring whether `x` is a pointer to a value, or a value.

Would a `Value` matcher be useful as a generic Gomega matcher, where it resolves any `*v` to its `v`, while passing a `v` as is?

onsi:
This makes sense to me.  I'm not sure that `Value` is the right name for the matcher - perhaps `ResolveValue`?

Also, how do you think should things behave if the pointer points to a pointer?  Does it keep resolving until it gets to a value?  Or does it stop after the first resolution?

TheDiveO:
- pointers to pointers: very good question.
  1. I wonder, if there is a real-world use case for a pointer value pointing to just a pointer value of its own type, outside `unsafe` tricks? I don't think so.
  2. "deep pointers", such as `**string` -- I think I've even seen triple-star'd code in C, but this looked very suspicious. However, I can see places where a double star might come in, albeit that might be C-style "out" parameter passing shoe-horned into Go. Do you think that's a real use case? A repeated pointer resolving won't be a problem and could be limited.
- naming: you mirror my thinking 😀. However, I'm under the impression that `Expect(pv).To(ResolveValue(Equal(42)))` stutters more than `Expect(pv).To(Value(Equal(42)))`. A con for `Value` is a higher chance for conflicting with the dev's types when using the usual dot-import. Do you think `HaveValue` might be better, such as in `Expect(pv).To(HaveValue(Equal(42)))`?

onsi:
re naming - `HaveValue` seems right to me.  `Expect(pv).to(HaveValue(Equal(42)))` conveys exactly what the user is intending to do.

[...]

PR #485, opened 2021-11-23
Add HaveValue matcher

TheDiveO:
- adds `HaveValue` matcher (as discussed in #484) that optionally dereferences pointers and interfaces into values, even repeatedly, as needed.
- adds tests for `HaveValue` matcher.

### DIFF
--- a/matchers/have_value.go
+++ b/matchers/have_value.go
@@ -7,11 +7,16 @@ import (
 	"github.com/onsi/gomega/types"
 )
 
-// HavePoint applies the given matcher to the resulting value after optionally
-// resolving actual to the value it points to or its interface value, repeatedly
-// as necessary. It fails if a pointer or interface is nil. In contrast to
-// gstruct.PointTo, HaveValue does not expect actual to be a pointer but instead
-// also accepts non-pointer and interface values.
+// HaveValue applies the given matcher to the value of actual, optionally and
+// repeatedly dereferencing pointers or taking the concrete value of interfaces.
+// Thus, the matcher will always be applied to non-pointer and non-interface
+// values only. HaveValue will fail with an error if a pointer or interface is
+// nil. It will also fail for more than 31 pointer or interface dereferences to
+// guard against mistakenly applying it to arbitrarily deep linked pointers.
+//
+// HaveValue differs from gstruct.PointTo in that it does not expect actual to
+// be a pointer (as gstruct.PointTo does) but instead also accepts non-pointer
+// and even interface values.
 //
 //   actual := 42
 //   Expect(actual).To(HaveValue(42))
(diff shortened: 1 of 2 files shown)


onsi:
hey @TheDiveO sorry for the delay (🦃-day and all).  This looks good but I'm gonna make a couple comments on the commit.

matchers/have_value.go
```diff
@@ -0,0 +1,78 @@
...
+	m.failure = format.Message(actual, "indirecting too many times")
+	return false, nil
+}
+
+func (m *HaveValueMatcher) FailureMessage(_ interface{}) (message string) {
+	return m.failure
+}
+
+func (m *HaveValueMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	return m.Matcher.NegatedFailureMessage(actual)
```
onsi, line 77: this line will likely cause some confusing failure messages.  how about rather than caching the failure message we cache the resolved actual value and pass it in to `FailureMessage` and `NegatedFailureMessage`?

matchers/have_value.go
```diff
@@ -0,0 +1,78 @@
...
+			elem := val.Interface()
+			match, err := m.Matcher.Match(elem)
+			if !match {
+				m.failure = m.Matcher.FailureMessage(elem)
+			}
+			return match, err
+		}
+	}
+	// too many indirections: extreme star gazing, indeed...?
+	m.failure = format.Message(actual, "indirecting too many times")
```
onsi, line 68: IMO this should be an error returned by `HaveValue` not a failure ('cause then `Expect(stargazer).ShouldNot(HaveValue(Equal(41))` will pass.)

matchers/have_value.go
TheDiveO: Ah, good catch! I took this from `PointTo`, but getting the original pointer or interface usually won't be that useful, except when trying to find solace in address bit patterns.

TheDiveO:
After finally being able to dynamically create arbitrary deep stargazing actual values using reflection, what do you prefer? Augmenting the PR with new commits or replacing the commits (if latter, how so)?

matchers/have_value.go
TheDiveO: Ahh, took me some time to finally grasp the important difference!

matchers/have_value.go
TheDiveO: How do I deal with the "not to be <nil>" failures? Are these actually failure or errors instead? `PointTo` sees them as failures, but in light of your comment on line 68 should this situation rather be an error, here and with `PointTo`?

matchers/have_value.go
```diff
@@ -28,51 +31,44 @@ func HaveValue(matcher types.GomegaMatcher) types.GomegaMatcher {
...
 	}
 	// too many indirections: extreme star gazing, indeed...?
-	m.failure = format.Message(actual, "indirecting too many times")
-	return false, nil
+	return false, errors.New(format.Message(actual, "too many indirections"))
 }
 
 func (m *HaveValueMatcher) FailureMessage(_ interface{}) (message string) {
-	return m.failure
+	return m.Matcher.NegatedFailureMessage(m.resolvedActual)
```
onsi, line 69: all looks great now!  just this one line - you want `m.Matcher.FailureMessage` not `m.Matcher.NegatedFailureMessage` (sorry if I somehow implied that earlier - if so I miscommunicated)

onsi: once this is tweaked I can pull this in - all else LGTM!

TheDiveO: That is something I obviously botches, so many thanks for spotting it. I'll have to think about a test to properly catch these types of mistakes.

TheDiveO: fixed.

onsi:
lgtm - thanks!

I have a couple of new things I want to add to Gomega this/next week - once that's done I'll cut a new release.

PR #491, opened 2022-01-06
Remove Travis workflow

williammartin:
Since the project moved to GitHub actions, this doesn't look like it should be used anymore. Ginkgo removed this a while ago.


onsi:
thanks!

PR #492, opened 2022-01-17
fix: use ioutil for Go 1.15 and lower

blgm:
Since Go 1.16 it has been recommended to no longer use ioutil in new
code. We want Gomega to look modern so we removed references to ioutil,
which also removes the chance of outdated code being copied. However we
got feedback that some users were stuck on Go 1.15 and lower and were
broken by this change. We have therefore introduces a "gutil" package
that implements similar functions to ioutil and redirect to the right
functions appropriate to the version of Go. This will hopefully:
- make it look intentional that ioutil is still used (rather than
looking like an omission)
- reduce the chance of ioutil usage being propagated by copying
- limit the possibility of deprecation warnings

### DIFF
--- a/gexec/build.go
+++ b/gexec/build.go
@@ -13,6 +13,8 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+
+	"github.com/onsi/gomega/internal/gutil"
 )
 
 var (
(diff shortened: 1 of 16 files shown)


onsi:
hey @blgm this looks good to me!  It's a fair compromise and I'm happy to pull it in.

Issue #520, opened 2022-02-17
[RFC] `HaveEach` matcher


thediveo:
In the hope that I didn't overlooked this functionality in the existing matchers: I would like to propose an `Each` matcher that checks that all elements of an array or slice match the specified value/matcher. Similar for maps, `Each` checks that all map values match the specified value/matcher.

```go
Expect(m).To(Each(Equal(42))) // map[string]int
Expect(m).To(Each(Each(Equal(42))) // map[string][]int
```

border case: `Expect([]string{}).To(Each(Equal("")))` is expected to succeed. If not desired, use this idiom:

```go
Expect([]string{}).To(And(Not(BeEmpty()), Each(Equal(""))))
```

I can work on a PR if you like and after some clean-up and adding the necessary tests :grin:.

P.S.: still struggling with the language semantics of `Each` and `Every`, but then Javascript used `forEach(...)` and not `forEvery(...)` :wink:

TheDiveO:
wouldn't `HaveAnyElement()` be the same as `ContainsElement()`?

onsi:
um... <awkward silence>... yes

TheDiveO:
Pondering again on `HaveAllElements()`  this appears to me ambiguous in the sense of "have all the following elements". Yet, you raise the import concern of name clashes with `Each`. What about `HaveEach()`? Avoids name clashes, keeps with the established `Have...` pattern, yet avoids the false friend of `ContainsElements()`. What do you think?

onsi:
Yeah, you're right. `HaveEach()` works and makes sense to me.

TheDiveO:
With this merged in #523 and #524 this issue is done. Thank you for the discussing the idea and merging!

PR #524, opened 2022-02-28
make HaveEach error in case of nil/empty actual val

TheDiveO:
relates to issue #520 and defines the HaveEach matcher behavior for empty arrays/slices/maps.

- HaveEachMatcher errors for empty actual value
- update unit tests
- update documentation and include example when empty actuals are acceptable


onsi:
Makes sense, good catch!

Issue #525, opened 2022-03-01
[RFC] "Expecting" ContainElement to return its findings


TheDiveO:
In some of my projects I end up far too often with some slice (such as `[]*Container`) where I expect one particular element to be contained in the slice. In addition, I then need to check further with several, more complex, expectations about that particular container. As lumping all together into a single long-winded matcher "freight train" isn't neither desired nor easy to understand and maintain, I usually ended up in the past with:

```go
var cntrs []*Container = someFoo()
var container *Container
for _, c := range cntrs {
    if c.Name == "foo" && c.Type == "bar" {
        container = c
        break
    }
}
Expect(c).NotTo(BeNil())
// ...more Expect'ations
```

After the fifth time or so this really gets tiresome and error-prone, so I wondered why not to reuse all the beautiful Gomega matchers and have a `ContainElement`-like filter that accepts the usual match, iterates over its actual (as usual), but now _returns_ its findings, with me `Expect()ing` to get some finding(s) returned:

```go
var cntrs []*Container = someFoo()
var container *Container
Expect(cntrs).To(ReturnElement(
        And(HaveField("Name", "foo"), HaveField("Type", "bar")).
    In(&container))
// ...more Expect'ations
```

This returns the matching elements in `container` and expects exactly one find/match -- because `*Container` points to a scalar, not a slice. In contrast, if passing the address of a slice, at least one find/match would be expected.

The naming might need some improvement. :grin:

And I freely admit, that this is taking Gomega into new territory.

What do you think? If you're interested, I can make a PR to better discuss it. I've already a working implementation with unit tests, but would like to further discuss this crazy idea in order to improve it.

onsi:
hey @TheDiveO sorry for the delay.  I've been heads down on a different project these days but I'm back now :)

> And I freely admit, that this is taking Gomega into new territory.

_Actually..._ not [quite](https://pkg.go.dev/github.com/onsi/gomega#Receive).

I'm a fan of the idea.  We could go with `ReturnElement(<matcher>).In(<pointer>)`  as you suggest but I think I'd sooner have the signature be `ReturnElement(<matcher>, <pointer>)` as I imagine it's an error (or at least, a typo) to not include the `.In()` part.

[...]

PR #527, opened 2022-03-16
Feature/525 return findings for ContainElement

TheDiveO:
implements RFC #525:
- implements an `optionalPointer` parameter to `ContainElement` for returning matching contained element(s)
- adds unit tests related to returning matching contained elements
- updates documentation for `ContainElement`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -886,12 +886,42 @@ succeeds if the capacity of `ACTUAL` is `INT`. `ACTUAL` must be of type `array`,
 Ω(ACTUAL).Should(ContainElement(ELEMENT))
 ```
 
+or
+
+```go
+Ω(ACTUAL).Should(ContainElement(ELEMENT, <Pointer>))
+```
+
+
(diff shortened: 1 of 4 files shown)


TheDiveO:
One note: the diff for the unit tests look ugly, but that's an unfortunate flaw of beloved diff. I've checked again and what I did was to wrap the existing tests for `ContainElement` in a `Describe` for the non-result-returning cases and then added separate new unit tests that cover the cases of returning macthed contained elements. More tests on the correct and useful error handling when the optional return pointer points at something that doesn't fit in with the type of actual.

onsi:
hey @TheDiveO how would you feel about becoming a collaborator with commit rights?  you're contributions have been super valuable and thoughtful!

No expectations on my end to do any maintenance/issues/PRs or anything like that (though if you're interested I won't say no 😉) - and I'm always happy to work on RFCs/PRs with you - but I don't want to be a blocker when my schedule is constrained.

TheDiveO:
Hi @onsi I don't mind about the delays, they're part of our lives. I feel very honored by your offer and gladly accept, albeit I still want PR feedback and discussions, except for maybe some minor type fix in the documentation as I happen to come across them.

onsi:
🎉 sounds great!  will add you now 😊

PR #535, opened 2022-03-25
Use latest ginkgo

BooleanCat:
The fun doesn't stop - this change slims down the `go.sum` significantly because Go 1.18 deals with the circular imports better.

I'll leave this PR in draft until Ginkgo is shipped, at which point I'll point it at the new version.


BooleanCat:
Sweet, this is good to go.

onsi:
🎉

Issue #536, opened 2022-04-01
Feature: goroutine leak testing


thediveo:
This moves further discussion on a new `gleak` experimental/beta package for goroutine leak detecting and matching from [Ginkgo issue 603](https://github.com/onsi/ginkgo/issues/603) to Gomega, as discussed. A corresponding PR will follow.

thediveo:
So I found a goroutine leak in one of my modules ... in some of the _tests_. 👀

Turns out that pulling container images using the containerd client leaves a persistent HTTP connection and some goroutines associated with it in the background. Fixing it by passing my own `http.Client` and then `defer`-ing closing all that HTTP client's idle connections. Something new learnt.

thediveo:
@onsi while hunting my leaks I noticed (not unexpectedly) some inconvenience in the leak reporting. I would like to hear your recommendations based on your experience. Originally, I run into two issues: 1. Gomega quickly truncating the overall matcher failure Object dump of the `Goroutines`s, because the backtraces quickly add up to the default 4k limit. 2. this usually truncates off also several of the goroutines which usually are important.

At the moment, [`FailureMessage`](https://github.com/thediveo/noleak/blob/f2953eca05b7f175d365d31cde55f278e747c1a1/have_leaked_matcher.go#L176) uses a hand-rolled message instead of Gomega's `format.Object`. [`listGoroutines`](https://github.com/thediveo/noleak/blob/f2953eca05b7f175d365d31cde55f278e747c1a1/have_leaked_matcher.go#L187) generates the Goroutine dump, but currently leaves out the backtrace. However, the backtraces are often crucial in understanding what really is going on, as the topmost function name and creator of the goroutine often don't suffice.

What would you suggest how to generate (format) the goroutines with backtraces? I would like to not have any truncation. Would leaving out the file names and line numbers good or bad? Any other ideas?

onsi:
hey @thediveo - it's super cool to see some immediate benefits from using this in your existing test suites :)

[...]

PR #538, opened 2022-04-04
implements new gleak goroutine leak detection experimental package

thediveo:
- implements issue #536; the code and documentation is under Gomega's license.
- provides goroutine discovery as goroutine.Goroutine
- provides HaveLeaked matcher for reasoning on slices of goroutines for leak detection
- provides set of IgnoringXXX matchers for identifying good (non-leaky) goroutines
- provides unit tests for gleak and gleak/goroutine packages
- updates documentation and adds Gleakiee, the leaky gopher, mascot


docs/index.md
```diff
@@ -2951,4 +2951,292 @@ Describe("server performance", func() {
 })
 ```
 
+## `gleak`: Finding Leaked Goroutines
```
onsi, line 2954: These docs are *fantastic* ! Thanks for taking the time to write them up!

thediveo:
No. (Except to duck down, maybe) The irony of it is that now I'm finding goroutine leaks in my _tests_, but the production code seems to be fine.

I'm wondering if we should come up next with a file descriptor leak tester, albeit I only know how to do this on Linux thanks to its `procfs` file system that lists the open fds of a process.

onsi:
oh man file descriptor leaks are the worst.

This all looks great, gonna merge it in now!

PR #539, opened 2022-04-08
document how to wrap Gomega

tzneal:
Add docs on how to use the new `inner` interface and why it may be useful.


onsi:
Hey this is great!  Thanks for putting together the PRs :)

PR #544, opened 2022-04-17
support pointer receivers in HaveField; fixes #543

### LINKED ISSUE #543: `HaveField` fails to find pointer receiver method
`HaveField` only locates non-pointer receiver functions, but not pointer receivers. In the following [minimal example](https://go.dev/play/p/H7bmNwDXvKJ) (on playground), `HaveField` locates the non-pointer receiver for `T`, but then fails for the pointer receiver of `T2`.

```go
package main

import . "github.com/onsi/gomega"

type T struct {
	s string
}

func (t T) S() string { return t.s }

type T2 struct {
	s string
}

func (t *T2) S() string { return t.s }

func main() {
	RegisterFailHandler(func(s string, cs ...int) { println(s) })
	var i interface{}

	i = &T{s: "D'OH!"}
	Expect(i).To(HaveField("S()", Equal("D'OH!")))

	i = &T2{s: "D'OH!"}
	Expect(i).To(HaveField("S()", Equal("D'OH!")))
}
```

I would have expected `HaveField` to shield test writers from such "nitty" details. Should we fix it (the matcher, not the writer)? How so; by taking the pointer to the struct value when the lookup fails and retrying again?

thediveo:


thediveo:
This might need some discussion? In case the receiver could not be found in the first lookup, I take the actual value's address and give it a second try.

thediveo:
Things in the mirror ... they say, and that means reflect(ion).

Issue #264, opened 2018-02-05
Failed comparison of identical time variables


wolfgangmeyers:
When I try comparing two time variables in a test:

```
	It("Should show identical times as equal", func() {
		now := time.Now()
		data, _ := now.MarshalJSON()
		copy := time.Time{}
		copy.UnmarshalJSON(data)
		Expect(now).To(Equal(copy))
	})
```

It fails with output like this:
```
  Expected
      <time.Time>: 2018-02-05T14:03:52.391613-08:00
  to equal
      <time.Time>: 2018-02-05T14:03:52.391613-08:00
```

I think that the `Equal` matcher should identify the two time variables as being equal.

nodo:
Thanks @wolfgangmeyers !

I was able to reproduce the problem. I have also tried to add some Printf:
```
fmt.Printf("%v\n", now)                                                                                                                             
fmt.Printf("%v\n", copy)
```

And this is the result:
```
2018-02-07 09:20:36.429395819 +0000 GMT
2018-02-07 09:20:36.429395819 +0000 UTC
```

It seems that `time.Location` for the two times is indeed different. There is also an issue open in `golang` which seems relevant (https://github.com/golang/go/issues/10089).

We could potentially catch that in `Equal` and instead of calling `reflect.DeepEqual` use https://golang.org/pkg/time/#Time.Equal .

williammartin:
Cool investigation @nodo. I sort of wonder whether there's use in having a separate set of time related matchers. I can definitely see how it would be nice to have `BeAfter`, `BeBefore` etc etc. Just food for thought.

onsi:
Yes this bites folks a lot.  The issue, as you've observed @nodo, is a mismatch in timezones.  This could be solved with a time matcher (I believe the little known `BeTemporally` has the right behavior?).  Updating `Equal` to do type-checking and use `Time.Equal` would help as well.....

...but the _real_ pain point is when you have mismatched time zones for time objects buried inside a struct.  A recent example of this bit a team at Pivotal where a struct contained a field of type `Time` and the deserialized JSON `actual` did not match the hand-constructed `expected`.  Without rewriting `reflect.DeepEqual` to type-check for `Time` and use `Time.Equal` I'm not sure we can solve for this case which is, often, even _more_ confusing that the example in this issue.

Any thoughts on what we can do here?

wolfgangmeyers:
In the example code I pasted, I believe the timezones are pacific for both the original and copy.

[...]

Issue #216, opened 2017-06-06
DeepEqual comparison should tell you what's not equal


benmoss:
Right now when you use `Expect(..).To(Equal(...))` and it uses go's DeepEqual function, it doesn't really help you find what is not equal. I just spent a bunch of time trying to figure out that an interface field was using a value type instead of a pointer.

Along the way I found https://godoc.org/github.com/juju/testing/checkers#DeepEqual, which is an alternate implementation of DeepEqual that actually will show you the differences. Would you consider a PR that changed gomega `Equal` to use this instead?

robdimsdale:
@benmoss would that introduce a new, third-party dependency to Gomega? If so, I foresee this causing pain for consumers of this library.

Is it better to re-implement that logic inside gomega than pull in a dependency?

onsi:
+1 to reimplementing.  This is dangerous territory... asserting equality can be tricky business in Go and relying on DeepEqual is a safe way to do so.  Maintaining control over the code would be key.

benmoss:
er, sure. i'm not sure if you're disagreeing with the original scary idea of reimplementing DeepEqual in a way that gives insight into what the inequality is, or if you're just scared of adding a dependency to gomega. The latter is solvable, the former not so much.

onsi:
My response wasn't very clear.  I would love to have improved diff identification in Gomega.  I'd want the code that does equality to be:

- a part of Gomega so we can be more responsive to issues
- forked off of Golang's [deep equal](https://golang.org/src/reflect/deepequal.go?s=5823:5860#L171)
- super well-tested.  I'd want to at least use [Go's own tests as a starting point](https://github.com/golang/go/blob/master/src/reflect/all_test.go#L667-L731)
- in it's own package under Gomega so it can be used as a standalone thing

Does that make sense?

benmoss:
Yup, I'm gonna try to work on this now!

benmoss:
ok after looking at this for a lil bit it seems way too hard to reimplement, my reflect-fu is far too lacking

benmoss:
I ended up making my own lib for this, in case anyone wants it: https://github.com/benmoss/matchers/

The burden of the criteria needed to be included in Gomega itself was just too high, so it's better to be a buyer-beware third party thing :).

alculquicondor:
Did you ever consider to depend on https://github.com/google/go-cmp?

alculquicondor:
A separate comparer sounds good. We have set it up like this for now:

[...]

PR #546, opened 2022-04-23
feat: add Compare using go-cmp

xiantank:
Add new matcher `Compare`, which use go-cmp to compare and we can pass cmp.Option as optional arguments.
here is example
```go
Expect([]int{1, 2}).Should(Compare([]int{1, 2}))
Expect(myCustomType{s: "foo", n: 3, f: 2.0, arr: []string{"a", "b"}}).Should(Compare(myCustomType{s: "foo", n: 3, f: 2.0, arr: []string{"a", "b"}}, cmp.AllowUnexported(myCustomType{})))
Expect(myCustomType{s: "abc", n: 3, f: 2.0, arr: []string{"a", "b"}}).Should(Compare(myCustomType{s: "foo", n: 3, f: 2.0, arr: []string{"a", "b"}}, cmpopts.IgnoreUnexported(myCustomType{})))
```

related: #216, #264


thediveo:
Just my tiny observation as I happened to look at go-cmp just these days (while working on checking file descriptor objects for equality). I remember that go-cmp thankfully documents its purposedly designed behavior of `panic`king, because it is intended to work without any matcher framework. It cannot rely on the usual error handling pattern that the Gomega matchers uses: matchers do not only return their "verdict" but also an error return value. I would, erm, expect that without handling panics the Compare matcher cannot be used in inverse expectations such as `Expect(foo).NotTo(Compare())`. If I didn't overlook something in your PR code (and please forgive me if I do and correct me please) then there's a inverse check using `Equal` but not one using `Compare`?

xiantank:
Hi, I add some panic handling to make Compare matcher as usual error handling pattern, and remove a weird inverse check Equal (which I tried to show difference between `Equal` and `Compare` deal with different time location, but it's really weird to do so, thanks for pointing it out).
// please let me know if I misunderstood anything

onsi:
hey @xiantank - can we rename it to `BeComparableTo`?  That would flow better for the DSL `Expect(foo).To(BeComparableTo(bar))`

I think I'd, eventually, like to make it possible to configure Gomega so that `Equal` can be backed by `go-cmp` instead of `reflect.DeepEqual`.  This is a great step in that direction, thanks!

onsi:
thanks! looks good to me!

onsi:
if you want to update the doc PR I can pull that in too

Issue #548, opened 2022-04-23
RFC: a "forgiving" HaveField matcher


thediveo:
In the past months, Gomega's matcher design has ventured -- or slightly gone astray 😬 -- into reusing matchers not only in a "direct" strict context, but also as filters. For instance, `HaveField` has on purpose "rigid" semantics: the tester expects `HaveField` to fail if the specified field is missing, not to silently not match. This clearly is important to quickly point out any mistakes in tests. `HaveField` is to reason about field values, not about field existence.

However, things get into broader scope when appropriating matchers for filtering instead: sometimes, the expectations are as before, such as "this field must be present". In other cases, such as with the recent experiments in resource leak detection for goroutines and file descriptors, a missing field isn't a failure but not a match instead.

While it might be that `gstruct` might already support such cases, especially `HaveField` has pathed the way for a more lightweight approach. I would like to ask for feedback and ideas how we can introduce matcher forgiveness in a way that doesn't water down the existing purposedly rigid error behavior, yet allows for lee room without introducing too much potential for hidden test bugs.

1. my gut feeling is that extending the existing `HaveField` matcher args would be feature creep, if possible at all. Or wouldn't it?
2. similar, I'm not yet seeing how a generic `Forgiving()` matcher might help that simply swallows any matcher errors, mapping them to non-matches instead. It looks like running with scissors to me. Do we even have the need for forgiveness when using other matchers?
3. what would you expect `HaveOptionalField` to default to if the field is missing? Matching or non-matching? Would make `HaveOptionalField(ACTUAL, VALUE, DEFAULT)` be making things better or worse?
4. ...or tackling this from a completely different angle, as I'm stuck in a corner with my idea?

onsi:
hey @thediveo - I think I'm struggling to get concrete with the application you have in mind for `HaveField` in particular.  Perhaps you could change tack and share an example problem you're trying to solve so I can better wrap my head around the question?

Absent that, I read this a few times and kept coming back to "is this solved with `SatisfyAny()`".  Like, `HaveOptionalField()` is really trying to model `SatisfyAny(HaveField("bloop", "blarp"), <Something Else>)`

thediveo:
[...]

PR #553, opened 2022-05-07
HaveExistingField matcher

thediveo:
- implements issue #548 with new HaveExistingField matcher
- modifies existing extractField helper from HaveField for reuse with HaveExistingField
- adds new unit tests for HaveExistingField matcher
- updates documentation

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -889,7 +889,7 @@ succeeds if the capacity of `ACTUAL` is `INT`. `ACTUAL` must be of type `array`,
 or
 
 ```go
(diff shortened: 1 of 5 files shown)


onsi:
whoops!  sorry for the delay.  I took a look and this makes sense.  The notion of a severe error in the implementation also makes sense.  An alternative could be to return an explicit FieldNotFoundError and key off of that in the HaveExistingField matcher but it's nbd.  we can merge as is imo

thediveo:
Your suggestion about changing from severe error to field-not-found absolutely makes sense and semantically much more so than the notion of severe error. I will refactor the PR, highly appreciated suggestion!

PR #558, opened 2022-06-12
[gleak] export Goroutine type alias for convenience; update documentation

thediveo:
Reduces the previous `goroutine.Goroutine` stuttering quite considerably.


onsi:
looks great!

PR #563, opened 2022-07-16
fix: pass cmp.Options to cmp.Diff in BeComparableToMatcher

bartlomiej-grzegorek:
Small fix to [this PR](https://github.com/onsi/gomega/pull/546) introduced @xiantank.
As additional `cmp.Options` might be passed to `cmp.Equal` method, the same should be passed to `cmp.Diff`. 
In worst cases, lacking of options might cause a panic; eg. while using
```golang
type structWithUnexportedFields struct {
    unexported string
    Exported   string
}

actual := structWithUnexportedFields{unexported: "xxx", Exported: "exported field value"}
expectedEqual := structWithUnexportedFields{unexported: "yyy", Exported: "exported field value"}
Expect(actual).To(BeComparableTo(expectedEqual, cmpopts.IgnoreUnexported(structWithUnexportedFields{}))
```
If the exported values were different, `cmp.Equal` returns `false` and `FailureMessage` will panic because `cmp.Diff` can't handle unexported fields.

### DIFF
--- a/matchers/be_comparable_to_matcher.go
+++ b/matchers/be_comparable_to_matcher.go
@@ -3,6 +3,7 @@ package matchers
 import (
 	"bytes"
 	"fmt"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/onsi/gomega/format"
 )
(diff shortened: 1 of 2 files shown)


onsi:
thanks for this - makes sense.  I'll pull it in once the GitHub actions go green.

Issue #560, opened 2022-06-16
Test should fail if extra matchers are given like Should(matcher, extraMatcher1, ...)


eiiches:
Currently, the extra matchers given to `Should(matcher, extraMatcher1, ...)` are silently ignored and the assertion passes successfully (see the example below). This happens when a user forget (...me, actually) to wrap the matchers with SatisfyAll() or other composition matchers. The problem is that this gives the wrong impression to the user that the SUT is working correctly.

It would be great if Gomega can detect this kind of incorrect usage and fail the test, notifying the user that the test should be fixed.

Example: (Gomega 1.19.0)

```go
package main

import (
  . "github.com/onsi/gomega"
  "testing"
)

func TestA(t *testing.T) {
  g := NewWithT(t)
  g.Expect([]int{1}).Should( // missing SatisfyAll
    HaveLen(1),
    ContainElement(2),
  )
}
```
```sh
$ go test .
ok      <import path redacted>    0.003s
```

blgm:
```go
Should(matcher types.GomegaMatcher, optionalDescription ...interface{})
```

The issue is that the `optionalDescription ...interface{}` is not validated unless the assertion fails. It seems reasonable that the length and type could always be validated so that usability issues such as the one outlined above could be improved.

thediveo:
Can work on a PR; should having a Gomega matcher in the `optionalDescription` cause a panic or a fail? I would expect the latter, as this is what we do when vetting the actual values. However, optionalDescription is part of the static test, so maybe a panic would be better?

thediveo:
This is fixed on Master. Please feel free to reopen if necessary.

PR #566, opened 2022-07-25
vet optional description args in assertions, fixing #560

thediveo:
- panics when it detects a Gomega Matcher in the first position of optional description args.
- accepts Gomega Matchers as part of an optional description when in other positions as args to the formatted first description arg.
- comes with unit tests.


onsi:
looks great!!

Issue #555, opened 2022-05-25
AsyncAssertion can't handle Eventually(functionReturnsNilError, <interval>, <interval>).Should(BeNil())


kaovilai:
It will panic when actualType.Kind() gets called
https://github.com/onsi/gomega/blob/fd130e126467fbdee60e707e887df557f3e642cf/internal/async_assertion.go#L43

Failing sample parameter to function.
actualType is AsyncAssertionType(0)
actualInput is nil <interface {}>

example function
```
func returnsNilError() error {
    return nil
}
```

thediveo:
Can you please show the code that you are using with `Eventually(???)` and especially what you are passing into `Eventually`, as well as the definition of your function you're passing? Could it be that you are passing the result of calling your function instead of passing the function? ...because you say that your test panics in async_assertion.go, line 43 ... and this is when creating the async assertion and thus `Eventually(nil, ...)`.

PS: it would be helpful if you could report the original panic trace in such situations, as it gives us a much better picture, especially with respect to the parameters (including the offending `actualInput` arg, respectively `actual`) a few levels deeper down the stack.

kaovilai:
You maybe right about the passing va LL he instead of function. If I get the time to try again I'll report here.

Still panicking should be avoided. You should be able to use defer recover pattern to give a more helpful error.

thediveo:
@onsi This actually raises the question of how we should handle a "constant" `nil` value. Is passing in a constant `nil` actually a valid use case or should we consider it to be a structural test fault where we should panic with a clear description as to why a permanent `nil` is considered to be a sign of the test programming making a mistake?

thediveo:
This should be fixed now and passing in a `nil` value won't cause a panic but also not warn you about probably doing the wrong thing; this is in line with passing in other non-`func` values. If we should still please miss something, please feel free to reopen and describe in detail what's missing. Thank you very much for reporting this bug!

kaovilai:
Thanks @thediveo !

PR #567, opened 2022-07-26
don't panic on Eventually(nil), fixing #555

thediveo:
- treat a `nil` actual value to `Eventually` as any other value that isn't a `func`.
- adds unit tests.


onsi:
hey @thediveo  - sorry I never managed to circle back on #555 - this seems like the correct fix to me and puts `nil` on par with other non-function values.

thediveo:
"circling back" evokes images of Anden condors...

onsi:
😂

PR #594, opened 2022-09-21
Adding simple ability for multiple responses from a test server

KevinJCross:


onsi:
hey thanks for this - i'll take a look when i'm at my computer :)

thediveo:
@KevinJCross would you mind also provide a few bits of information and example usage for the documentation? -- the https://onsi.github.io/gomega/ one; you already documented the source code.

KevinJCross:
hi @onsi and @thediveo 
So there is documentation how to use this in the test examples which should show up in the godocs. I.E. the function  ExampleRespondWithMultiple should show up in the godocs its also a test btw.

The problem is that append handlers and Set default handler does not add a "use case" the normal use case for this is for testing retries and creating a stress test with some problematic behaviour. The adding of the ResponseWithMultiple is obvious in the way it has one set of circular handlers for **one** single endpoint.

In our case we set up a whole environment of services and responses on the single ghttp server so setting the default response to be repeated could be confusing.

Most of the time Ive seen ghttp been used for single request validation
i.e:
```go
                             server.AppendHandlers(
					CombineHandlers(
						VerifyRequest("GET", "/v3/something/404"),
						RespondWithJSONEncoded(http.StatusNotFound, cf.CfResourceNotFound),
					),
				)
```

An example of using this to test some retry behaviour:
	https://github.com/cloudfoundry/app-autoscaler-release/blob/bb426fb82aa9ea081984b0bcb98d24e20a4d6611/src/autoscaler/cf/retriever_test.go#L76-93

An example of using this to stress test some failure and recovery behaviour (I deved this test to find a subtle behaviour that was plaguing one of our deployments), it is a bit complicated but I had to have a repeating behaviour on one sepcific end point an have this endpoint hit by several "threads"/users at the same time.
https://github.com/cloudfoundry/app-autoscaler-release/blob/bb426fb82aa9ea081984b0bcb98d24e20a4d6611/src/autoscaler/cf/app_test.go#L188-L248

Ive used other mocking tools were you can create a "scene" or "act" (like a play) where its easy to just write a whole bunch of request responses in sequence to elicit the incorrect behavour you are trying to fix. I think this is little easier to read in sequence than the above situations.

Maybe ghttp is the wrong tool for this but it was what was in hand without having to add more dependencies

PR #627, opened 2023-01-29
ci(lint): setup lint job

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,19 @@ name: test
 on: [push, pull_request]
 
 jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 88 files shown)


onsi:
hey there @mmorel-35  - thanks for wrangling those other dependency-relate updates.

I'm happy with the current state of `go fmt`ing of the codebase, however, and will not be pulling this PR in

mmorel-35:
Hi @onsi ,

I can create a PR with gofmt instead of gofumpt if you are interested in golangci. There won't be as much modification with it.
Would you like that ?

onsi:
Thanks for the offer, but I don't think so.  It's just not a usecase that I have a lot of need for.

PR #634, opened 2023-02-01
Add HaveExactElements matcher

### LINKED ISSUE #622: Is there any support for matching ordered slice?
Hi! Thanks for the awesome matcher package!

I know that currently we can use `Equal` and friends to have exact matching of ordered slice.
But what if I want to match the slice, in an ordered manner, with custom equality matcher?

Something like this
```go
tc := []string{"aaa", "bbb", "ccc"}
Expect(tc).To(ConsistOf(
  ContainsSubstring("a"), // "aaa"
  Equal("bbb"), // "bbb"
  BeEquivalentTo("ccc"), // "ccc"
))
```

My current use case is that I have list of proto message which I can't use exact equality. I'm using the [gproto](https://github.com/jt-nti/gproto) package to do the equality. Something similar to the snippet below.
```go
tc := []proto.Message{...}
expectation := []proto.Message{...}
matchers := []types.GomegaMatcher{}
for _, ex := range expectations {
  matchers := append(matchers, gproto.Equal(ex))
}
Expect(tc).To(ConsistOf(matchers)) // but no ordering
```

Currently, I'm using loop to match the index with matcher's index to get the ordering. Is it expected that way or is there something I am missing?

rickyson96:
This is the PR for implementing the ordered matcher.

Closes #622

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1215,6 +1215,38 @@ is the only element passed in to `ConsistOf`:
 
 Note that Go's type system does not allow you to write this as `ConsistOf([]string{"FooBar", "Foo"}...)` as `[]string` and `[]interface{}` are different types - hence the need for this special rule.
 
+#### HaveExactElements(element ...interface{})
+
+```go
+Expect(ACTUAL).To(HaveExactElements(ELEMENT1, ELEMENT2, ELEMENT3, ...))
+```
+
+or
+
+```go
+Expect(ACTUAL).To(HaveExactElements([]SOME_TYPE{ELEMENT1, ELEMENT2, ELEMENT3, ...}))
+```
+
+succeeds if `ACTUAL` contains precisely the elements and ordering passed into the matchers.
+
+By default `HaveExactElements()` uses `Equal()` to match the elements, however custom matchers can be passed in instead.  Here are some examples:
+
+```go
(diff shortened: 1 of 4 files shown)


rickyson96:
I wanted to implement some mechanism of reporting the matching failure.
Something like this
```
Expected
    <[]int | len:2, cap:2>: [1, 2]
to have exact elements with
    <[]int | len:2, cap:2>: [2, 1]
the mismatch indexes were:
    index: 0, actual: 1, expects: to equal 2
    index: 1, actual: 2, expects: to equal 1
```

I'm thinking of utilizing the matcher messages used to fail the elements, but don't want to have the newlines.
Is there anyway to do that? Or maybe other alternative solution?

Thanks! 😄

onsi:
hey! sorry for the delay - thanks for pulling this together.  i'll take a look tomorrow and share feedback/pull it in.  for now - it looks like one of the tests failed on ci?

rickyson96:
Hi, sorry for the late response.

I have updated the implementation on the test reporting.
I have also fixed the failed test, seems like I accidentaly ran `ginkgo bootstrap` there

onsi:
Looks great, thanks - I'll merge this in after the tests go green.  Would you be up for adding a brief snippet of documentation to docs/index.md ? Fine if not, I can do it after I pull it in.

Thanks!

rickyson96:
done! 😁

PR #648, opened 2023-03-06
Fix HaveExactElements to work inside ContainElement or other collection matchers

### LINKED ISSUE #647: HaveExactElements() doesn't work properly inside ContainElement or collection matchers
Hi,

I've encountered a weird behavior when using HaveExactElements matcher. See this example:

```go
actual := [][]bool{
	{true},
	{false},
}
g.Expect(actual).To(ContainElement( // this fails, even though it should pass
	HaveExactElements(Equal(false)),
))
```

I expect this assertion to pass, because `actual` contains an element `[]bool{false}` which matches `HaveExactElements(Equal(false))`. But the assertion fails instead.

I believe the culprit is that HaveExactElementsMatcher leaves some internal state once its Match() returns false, which never gets cleaned and causes subsequent calls to always return false. Fwiw, cleaning the state [at the beginning of the Match()](https://github.com/onsi/gomega/blob/dbd4cb504ac2e8b5cdbeb204e9c75447f56cc39a/matchers/have_exact_elements.go#L21-L22) fixed the issue for me.

```diff
 func (matcher *HaveExactElementsMatcher) Match(actual interface{}) (success bool, err error) {
+ 	matcher.missingIndex = 0
+ 	matcher.extraIndex = 0
+ 	matcher.mismatchFailures = nil
+ 
 	if isMap(actual) {
```

Version: 1.27.2

eiiches:
Fixes #647

As described in #647, HaveExactElements() leaves some internal state once its Match() returns false. The state never gets cleaned afterwards and causes subsequent calls on the instance to return incorrect results. This can be a problem when the matcher is placed inside ContainElement() or any other matchers that reuse and call inner matcher instances multiple times.

This PR fixes the issue by resetting the matcher state at the beginning of the HaveExactElements#Match().

### DIFF
--- a/matchers/have_exact_elements.go
+++ b/matchers/have_exact_elements.go
@@ -19,6 +19,8 @@ type HaveExactElementsMatcher struct {
 }
 
 func (matcher *HaveExactElementsMatcher) Match(actual interface{}) (success bool, err error) {
+	matcher.resetState()
+
 	if isMap(actual) {
 		return false, fmt.Errorf("error")
 	}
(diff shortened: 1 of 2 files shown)


onsi:
LGTM!  will merge when CI goes green

Issue #642, opened 2023-02-22
error checking: accept Error() as alternative to String()


pohly:
In Kubernetes, we (not often) check for specific errors:
```
              gomega.Expect(err).To(gomega.HaveOccurred())
              gomega.Expect(err.Error()).To(gomega.ContainSubstring(`Invalid value: "foo-"`))
              gomega.Expect(err.Error()).To(gomega.ContainSubstring(`Invalid value: "bar.."`))
              gomega.Expect(err.Error()).NotTo(gomega.ContainSubstring(`safe-and-unsafe`))
              gomega.Expect(err.Error()).NotTo(gomega.ContainSubstring("kernel.shmmax"))
```

I can currently write this as two assertions:
```
              gomega.Expect(err).To(gomega.HaveOccurred())
              gomega.Expect(err.Error()).gomega.ContainSubstring(`Invalid value: "foo-"`),
			gomega.ContainSubstring(`Invalid value: "bar.."`),
			gomega.Not(gomega.ContainSubstring(`safe-and-unsafe`)),
			gomega.Not(gomega.ContainSubstring("kernel.shmmax")),
	      ))
```

I would like to reduce this to one assertion:
```
		gomega.Expect(err).To(gomega.SatisfyAll(
			gomega.HaveOccurred(),
			gomega.ContainSubstring(`Invalid value: "foo-"`),
			gomega.ContainSubstring(`Invalid value: "bar.."`),
			gomega.Not(gomega.ContainSubstring(`safe-and-unsafe`)),
			gomega.Not(gomega.ContainSubstring("kernel.shmmax")),
		))
```

This currently fails:

```
  [FAILED] ContainSubstring matcher requires a string or stringer.  Got:
      <*errors.errorString | 0xc00153ebb0>: {
          s: "Invalid value: \"foo-\"",
      }
```

I think would be consistent with the gomega design to allow `error` as an alternative for a `fmt.Stringer` and to call `Error()` to obtain a string when given an `error`.

The advantages are:
- in case of a failure, the full error object gets dumped
- IMHO clearer code (one statement about `err` instead of many)

onsi:
hey there - obviously I'm open to this but want to point out what currently exists first.  You _can_ do all this with:

```go
Expect(err).To(MatchError(SatisfyAll(
	ContainSubstring(`Invalid value: "foo-"`),
	ContainSubstring(`Invalid value: "bar.."`),
	Not(ContainSubstring(`safe-and-unsafe`)),
[...]

PR #654, opened 2023-03-25
updates MatchError godoc comment to also accept a Gomega matcher

thediveo:
- see also https://github.com/onsi/gomega/issues/642#issuecomment-1483307709


onsi:
thanks!

PR #678, opened 2023-06-28
chore: use String() instead of fmt.Sprintf

testwill:


format/format.go
```diff
@@ -302,7 +302,7 @@ func formatType(v reflect.Value) string {
 	case reflect.Map:
 		return fmt.Sprintf("%s | len:%d", v.Type(), v.Len())
 	default:
-		return fmt.Sprintf("%s", v.Type())
+		return  v.String()
 	}
```
blgm, line 306: These don't output the same thing:
```go
func main() {
	v := reflect.ValueOf(5)
	fmt.Println(fmt.Sprintf("%s", v.Type()))
	fmt.Println(v.String())
	fmt.Println(v.Type().String())
}
```
I agree that the `Sprintf()` is not needed, but I think the replacement should be `v.Type().String()` rather than `v.String()`.

blgm:
Thanks @testwill

Issue #682, opened 2023-07-18
HaveHTTPBody Does Not Work with Eventually


Kyle-Cooley:
The following code does not work as expected when the first attempt does not succeed:
```golang
Eventually(func(ctx SpecContext) (*http.Response, error) {
  req, err := http.NewRequestWithContext(ctx, http.MethodPost, "http://localhost:8080/test", strings.NewReader(query))
  req.Header.Add("Content-Type", "application/json")
  if err != nil {
    return nil, err
  }
  return http.DefaultClient.Do(req)
}).
  WithContext(ctx).
  WithPolling(5 * time.Second).
  WithTimeout(30 * time.Second).
  Should(HaveHTTPBody(MatchJSON(expectedResult)))
```

This appears to be caused by the use of a cached variable that is never updated after being set:
https://github.com/onsi/gomega/blob/master/matchers/have_http_body_matcher.go#L76

Kyle-Cooley:
Possible solution:
Store the pointer for the request and use that to determine if the cache should be used or not.

onsi:
hey @Kyle-Cooley - good catch; agreed on checking the pointer as you're proposing.  Would you be up for submitting a PR?

Kyle-Cooley:
Hello @onsi , yes, I will make a PR for it.

PR #686, opened 2023-07-25
Read Body for Newer Responses in HaveHTTPBodyMatcher

Kyle-Cooley:
Fixes https://github.com/onsi/gomega/issues/682


blgm:
Thanks @Kyle-Cooley!

PR #698, opened 2023-09-27
Add VerifyHost handler to ghttp

tlwr:
What
-----

Adds a new handler to ghttp to verify the Host field of http.Request

Why
----

Host is a field on the http.Request struct and it is useful to test that its value is correctly set.

One cannot rely on checking the header (ie by `VerifyHeader`) alone


onsi:
lgtm, thanks!

PR #725, opened 2024-01-19
Show how to import the format sub package

luistm:
Show how to import the format sub package.
I find this useful for developers that are less familiar with Go and Gomega.


onsi:
thanks!

Issue #470, opened 2021-09-16
Migrate to google.golang.org/protobuf


jtherin:
The module `github.com/golang/protobuf` as been superseded by the `google.golang.org/protobuf` module, which contains an updated and simplified API, support for protobuf reflection, and many other improvements.

I can make a PR if you want.

onsi:
Hey, thanks!  A PR would be great.

PR #490, opened 2021-12-29
470: Use google.golang.org/protobuf/proto

Akaame:
Commit message: Update to google.golang.org/protobuf/proto
This ipdates the imports
This is a huge leap and it comes with some necessary changes
simple_message.proto includes go_package option now.
and the message equality of two protobuf messages should be checked by proto.Equal.

Update yaml to v3 and golang/x/net to the next version

Fixes #470

PR #743, opened 2024-03-15
Migrate github.com/golang/protobuf to google.golang.org/protobuf

jbduncan:
Fixes https://github.com/onsi/gomega/pull/470.

This PR fixes the merge conflicts in https://github.com/onsi/gomega/pull/490 and uses the latest `protoc` and `protoc-gen-go` to fix `simple_message.proto` with command `protoc --go_out=ghttp/protobuf ghttp/protobuf/simple_message.proto`.


onsi:
hey - thanks.  just wanted to check (I'm not super familiar with the protobuf ecosystem so bear with me if this is a dumb question): can I safely assume this  is a backward compatible change?

Thanks!

jbduncan:
@onsi Good question! I'm just as unfamiliar with the protobuf ecosystem as you are, but I had to regenerate `ghttp/protobuf/simple_message.pb.go` with the latest `protoc` compiler to prevent Go compilation errors when passing it to `gomega.VerifyProtoRepresenting`, so I think it is incompatible, sadly.

It looks like it's possible to pass instances of old `*.pb.go` files like our `simple_message.pb.go` to `VerifyProtoRepresenting` with `VerifyProtoRepresenting(protoadapt.MessageV2Of(simpleMessage))`, but it will break existing code and I don't know what the recommended way forward for libraries like Gomega is.

jbduncan:
@onsi I think I've found a way to maintain source backwards compatibility (see [e61836c](https://github.com/onsi/gomega/pull/743/commits/e61836c77078e13ec7c0cfecd6ae66571e945b4b)).

On master (which uses `github.com/golang/protobuf v1.5.3`), `VerifyProtoRepresenting` accepts `proto.Message` which is an alias for `protoiface.MessageV1`. But as of this PR changing things to `google.golang.org/protobuf v1.33.0`, protobuf broke backwards compatibility by changing `proto.Message` to be an alias for `protoreflect.ProtoMessage`, which seems to be Protobuf's "message V2" API.

Therefore, I've changed `VerifyProtoRepresenting` to accept `protoiface.MessageV1`, which is structurally the same interface as what `proto.Message` was before. Also the new version of `simple_message.pb.go` implements it, so V2 message types are backwards-compatible too, thankfully.

I decided not to revert `simple_message.pb.go` back to its old implementation because it used the previous package `github.com/golang/protobuf`, and it didn't feel right to leave Gomega's tests stuck with that package given it's been deprecated.

Any thoughts? Otherwise, I think we're good to go.

onsi:
thanks @jbduncan - you're now the expert :wink:

I'll merge this in and cut a release.  If folks open up issues I'll tag you on them so we can figure out what to do together if need be.  Sound ok?

jbduncan:
> thanks @jbduncan - you're now the expert 😉
> 
> I'll merge this in and cut a release. If folks open up issues I'll tag you on them so we can figure out what to do together if need be. Sound ok?

@onsi Haha! I suppose I am! Yep, that's totally fine with me. Thanks for merging this in so quickly. 👍

Issue #765, opened 2024-06-12
bug in BipartiteGraph::createSLAPGuideLayers lead to violates concern


timmyb32r:
In algorithm 'Hopcroft–Karp', augmentation path should starts and ends with unmatched vertex

but into BipartiteGraph::createSLAPGuideLayers in last layer appended all nodes (actually it's handled by the same code, that any another layer)

I have quite simple test, which demonstrates the problem:

```golang
package bipartitegraph

import (
	"github.com/onsi/gomega/matchers/support/goraph/edge"
	"slices"
	"testing"
)

func buildEdgesArr(l, r []interface{}, edges edge.EdgeSet) []string {
	unpackArr := func(in []interface{}) []string {
		result := make([]string, 0, len(in))
		for _, el := range in {
			result = append(result, el.(string))
		}
		return result
	}

	vertexes := unpackArr(append(l, r...))

	result := make([]string, 0)
	for _, currEdge := range edges {
		result = append(result, vertexes[currEdge.Node1]+"-"+vertexes[currEdge.Node2])
	}
	return result
}

func expectedContains(t *testing.T, expected string, edges []string) {
	idx := slices.IndexFunc(edges, func(c string) bool { return c == expected })
	if idx == -1 {
		t.Fatalf("edges %v not contains expected: %s", edges, expected)
	}
}

func TestMaximumCardinalityMatch(t *testing.T) {
	edgesFunc := func(l, r interface{}) (bool, error) {
		ll := l.(string)
		rr := r.(string)

		type currEdge struct {
			l string
			r string
		}
		knownEdges := []currEdge{
			{"1", "A"},
			{"1", "B"},
			{"1", "C"},
			{"1", "D"},
			{"1", "E"},
			{"2", "A"},
			{"2", "D"},
			{"3", "B"},
			{"3", "D"},
			{"4", "B"},
			{"4", "D"},
			{"4", "E"},
			{"5", "A"},
		}

		for _, el := range knownEdges {
			if el.l == ll && el.r == rr {
				return true, nil
			}
		}
		return false, nil
	}

	leftPart := []interface{}{"1", "2", "3", "4", "5"}
	rightPart := []interface{}{"A", "B", "C", "D", "E"}

	bipartiteGraph, err := NewBipartiteGraph(
		leftPart,
		rightPart,
		edgesFunc,
	)
	if err != nil {
		t.Fatalf("NewBipartiteGraph returned error: %v", err)
	}
[...]

PR #766, opened 2024-06-12
issue_765 - fixed bug in Hopcroft-Karp algorithm

timmyb32r:
see https://github.com/onsi/gomega/issues/765


onsi:
hi - thanks for digging into this and providing a fix.  sorry for the delay but i'm going to need to find some time to understand this and pull it in (and rewrite the test into a ginkgo test to match the other tests in the package).  hope to get to it in the next few weeks.

onsi:
thanks!  I've pulled this in and updated the tests to match the existing test style

PR #770, opened 2024-07-12
Add RoundTripper method to ghttp.Server

### LINKED ISSUE #769: Helper to create a RoundTripper for ghttp.Server
# Summary

Currently we can use `server := ghttp.NewServer()` to create a test server, then use `server.URL()` or `server.Addr()` to point our clients to the test server.

In some situations code is not written in such a way where the URL can passed to a function, or perhaps paths should be conserved. It is common for libraries to be able to configure a `http.Client`, and therefore a `http.Transport`. The interface `RoundTripper` allows us to take a request, modify it in someway, and return the response/error

## Proposal

Add a new method to `ghttp.Server` which returns a `RoundTripper`. This can then be used to build http clients, or wrap existing transports (i.e. in `req`'s `WrapRoundTrip` [link](https://req.cool/docs/tutorial/middleware-for-client-and-transport/)).

Because we need to send a real request to our test server, we still require an actual transport. Therefore, the method takes a `RoundTripper`. If this is `nil` then we can use `http.DefaultTransport`.

For example:

```golang
func (s *Server) RoundTripper(rt http.RoundTripper) http.RoundTripper {
	if rt == nil {
		rt = http.DefaultTransport
	}
	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
		r.URL.Scheme = "http"
		r.URL.Host = s.Addr()
		return rt.RoundTrip(r)
	})
}

```

Where the helper `RoundTripperFunc` is as such. Note there is an open issue to add this to `net/http` in https://github.com/golang/go/issues/38479.

```golang
type RoundTripperFunc func (*http.Request) (*http.Response, error)

func (fn RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
	return fn(r)
}
```

## Example usage

```golang
var _ = Describe("A Sprockets Client", func() {
	var server *ghttp.Server
	var client *SprocketClient
	BeforeEach(func() {
		server = ghttp.NewServer()
		httpClient := http.Client{Transport: server.RoundTripper(nil)}
		client = NewSprocketClient(httpClient, "skywalker", "tk427")
	})

	AfterEach(func() {
		server.Close()
	})
})
```

Smirl:
# Summary

Fixes #769 

Add a new method `RoundTripper` which can be used to create `http.Client` or otherwise when testing http.

This is useful if you cannot access the URL directly, or do not want to change the path.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2424,6 +2424,33 @@ To bring it all together: there are three ways to instruct a `ghttp` server to h
 
 When a `ghttp` server receives a request it first checks against the set of handlers registered via `RouteToHandler` if there is no such handler it proceeds to pop an `AppendHandlers` handler off the stack, if the stack of ordered handlers is empty, it will check whether `GetAllowUnhandledRequests` returns `true` or `false`.  If `false` the test fails.  If `true`, a response is sent with whatever `GetUnhandledRequestStatusCode` returns.
 
+### Using a RoundTripper to route requests to the test Server
+
+So far you have seen examples of using `server.URL()` to get the string URL of the test server. This is ok if you are testing code where you can pass the URL. In some cases you might need to pass a `http.Client` or similar.
+
+You can use `server.RounderTripper(nil)` to create a `http.RounderTripper` which will redirect requests to the test server.
+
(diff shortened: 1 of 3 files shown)


onsi:
thanks!

Issue #774, opened 2024-07-29
1.34.0 cannot be used with golang 1.20


gibizer:
Bumping from 1.33.1 to 1.34.0 in a golang 1.20 project fails with:
```
/home/gibi/go/pkg/mod/github.com/onsi/gomega@v1.34.0/matchers/support/goraph/bipartitegraph/bipartitegraphmatching.go:4:2: package slices is not in GOROOT (/usr/lib/go-1.20/src/slices)

```
See also: https://github.com/openstack-k8s-operators/nova-operator/pull/837#issuecomment-2253749772

I think the fix in https://github.com/onsi/gomega/pull/766 introduced "slices" dependency that is not in the standard lib in golang 1.20

onsi:
hey there - I'll pull in the PR but need to point out that there is no LTS release for Gomega.  Gomega supports the latest two versions of Go in line with Go's own [release policy](https://go.dev/doc/devel/release) - so 1.20 is, formally, not "officially" supported any more.

gibizer:
Thansk for accepting the fix.

Of course an alternative fix would be to declares 1.21 as minimum  golang version in the go.mod file. https://github.com/onsi/gomega/blob/fa057b845528c9336265d6df77e21d827f268c7d/go.mod#L3 
But since it was still stating 1.20 I thought that the intention was to keep the compatibility. So I formulated my fix in a way to make the least disturbing change.

Thank you for your effort on gomega!

PR #775, opened 2024-07-29
Use slices from exp/slices to keep golang 1.20 compat

gibizer:
The #766 introduced dependency to slices but that module is not in the
standard lib in golang 1.20.

Closes: #774


liggitt:
go 1.20 is out of support, adding the dependency on an unstable x/exp module seems worse to me and is blocking some downstreams from updating (xref https://github.com/kubernetes/kubernetes/pull/126787#issuecomment-2311504429)

onsi:
hey @liggitt - happy to pull this out and cut a new release that also bumps the minimum Go version to 1.22.0 which is in keeping with the Ginko/Gomega last-2-go-versions policy

liggitt:
it looks like the use is super minimal... maybe we could just delete an item from a slice or find an item in a slice using a simple loop instead of stdlib slices or x/exp use...?

onsi:
I’ll take a look - but, kinda out of principle, i should be doing a better job of bumping the minimum Go version more consistently to help suss these things out and convey clearly to the community that Ginkgo/Gomega expects you to be on the latest go versions.

PR #794, opened 2024-11-06
Replace min/max helpers with built-in min/max

Juneezee:
We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max


Juneezee:
@onsi friendly ping :blush:

Juneezee:
@onsi friendly ping :blush:

onsi:
hey thanks, so for the delay.

Issue #795, opened 2024-11-10
Go1.23 first-class iterator support in certain collection-related matchers


thediveo:
Now that Go 1.23 comes with (push) iterators, after playing around with some of my own code I notice ending up with writing tests like this one...

    foos := []Bar
    for _, foo := range Bars() {
        foos = append(foos, foo)
    }
    Expect(foos).To(ConsistOf("foo", "bar", "baz"))

...instead of...

    Expect(Bars()).To(ConsistOf("foo", "bar", "baz"))

Do you think that first-class support would make sense (even if some work), where matchers that are currently handling collections for their actual value, like slice and map, in a near future additionally handle `iter.Seq[V]` and `iter.Seq2[K, V]` right out of the box?

- `seq.Iter` and `seq.Iter2` actuals, where only the V values are taken into consideration in case of `seq.Iter2`:
  - [ ] `BeEmpty` 
  - [ ] `HaveLen` ... hm, oddball?
  - [ ] `ContainElement` and `ContainElements`
  - [ ] `ConsistOf`
  - [ ] `HaveExactElements`
  - [ ] `HaveEach`
- only for `iter.Seq2` actuals
  - [ ] `HaveKey`
  - [ ] `HaveKeyWithValue`

The `go.mod` would be at most at 1.22, with the new iterator-awareness hidden behind a `go:build go1.23` build constraint; the challenge is rather how to pimp the existing matchers without ending up in two completely different code paths that need to be maintained at least until go 1.24 will be out. Now, it would be natural to refactor the existing collection-related matchers to use iterators ... dang!

Any(thing) missing? Any suggestions? Any groans? 😁

thediveo:
I'm now working on this and `ContainElement` has gotten iterator-awareness, including partial test case coverage. `ContainElement` is optimized in that it does not simply loop over all elements to create a slice and then fall back to work on this slice, but instead does the checking directly on the iterator yields.

Probably `HaveEach` can also be optimized as well, but some other matchers might need to use the intermediate slice.

onsi:
hey sorry for the radio silence on my end - this is fantastic!  thank you 😊

thediveo:
update: now only two collection-related matchers remain to be done, `ContainElements` and `ConsistOf` are the oddballs, in that they need to collect all elements produced by an actual iterator in order to then feed into the bisect algorithm.

thediveo:
[...]

PR #798, opened 2024-11-20
new: make collection-related matchers Go 1.23 iterator aware, resolves onsi/gomega#795

thediveo:
- new internal helper package for dealing with Go 1.23 iterators via reflection; for Go versions before 1.23 this package provides the same helper functions as stubs instead, shielding both the matchers code base as well as their tests from any code that otherwise would not build on pre-iterator versions. This allows to keep new iterator-related matcher code and associated tests inline, hopefully ensuring good maintainability.
- with the exception of ContainElements and ConsistOf, the other iterator-aware matchers do not need to go through producing all collection elements first in order to work on a slice of these elements. Instead, they directly work on the collection elements individually as their iterator produces them.
- BeEmpty: iter.Seq, iter.Seq2 w/ tests
- HaveLen: iter.Seq, iter.Seq2 w/ tests
- HaveEach: iter.Seq, iter.Seq2 w/ tests
- ContainElement: iter.Seq, iter.Seq2 w/ tests
- HaveExactElements: iter.Seq, iter.Seq2 w/ tests
- ContainElements: iter.Seq, iter.Seq2 w/ tests
- ConsistOf: iter.Seq, iter.Seq2 w/ test
- HaveKey: iter.Seq2 only w/ test
- HaveKeyWithValue: iter.Seq2 only w/ test
- updated documentation.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -690,6 +690,8 @@ A number of community-supported matchers have appeared as well.  A list is maint
 
 These docs only go over the positive assertion case (`Should`), the negative case (`ShouldNot`) is simply the negation of the positive case.  They also use the `Ω` notation, but - as mentioned above - the `Expect` notation is equivalent.
 
+When using Go toolchain of version 1.23 or later, certain matchers as documented below become iterator-aware, handling iterator functions with `iter.Seq` and `iter.Seq2`-like signatures as collections in the same way as array/slice/map.
+
 ### Asserting Equivalence
 
 #### Equal(expected interface{})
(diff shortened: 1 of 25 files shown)


onsi:
wonderful :) will take a close look today/tomorrow.  thanks for this!

thediveo:
Yes on all your points!

There are different ways to alleviate:
- Go's "make it explicit" attitude
  - either standard (x) lib helpers to reduce iters to slices or maps before passing the latter as actual,
  - Gomera `giter` convenience package using generics
- doing the slice/map under the hood, either partial or completely
- keeping the already iterated elements to provide better details up to the point of failure.

thediveo:
and yes, please cut a release if you please 😀

onsi:
done!  and makes sense on the ways to alleviate.  Thanks for this super valuable contribution!

Issue #696, opened 2023-09-06
HaveField: cannot call method with pointer receiver for values


pohly:
I [recently suggested](https://github.com/kubernetes/kubernetes/pull/119459#discussion_r1314758315) replacing these checks with `ConsistOf + HaveField("GetName()", Equal`:
```
	gomega.Expect(len(list.Items)).To(gomega.BeIdenticalTo(2))
	framework.ExpectEqual((list.Items[0].GetName() == name1 && list.Items[1].GetName() == name2) ||
		(list.Items[0].GetName() == name2 && list.Items[1].GetName() == name1), true)
```

=>

```
gomega.Expect(list.Items).To(gomega.ConsistOf(gomega.HaveField("GetName()", gomega.Equal(name1)), gomega.HaveField("GetName()", gomega.Equal(name2))))
```

That didn't work. It turned out that the slice elements are values and their `GetName` method is only defined for pointers. Go can call it anyway, but Gomega can't (or doesn't). Perhaps it should also support that?

Here's a reproducer:
```
package gomegatest

import (
	"testing"

	"github.com/onsi/gomega"
)

type pointerobject struct {
	name string
}

func (p *pointerobject) GetName() string {
	return p.name
}

type valueobject struct {
	name string
}

func (v valueobject) GetName() string {
	return v.name
}

type object interface {
	GetName() string
}

var _ object = valueobject{}

// Only pointer implements object interface, not value.
var _ object = &pointerobject{}

// var _ object = pointerobject{}

func TestConsistsOf(t *testing.T) {
	g := gomega.NewGomegaWithT(t)
	e := gomega.ConsistOf(gomega.HaveField("GetName()", gomega.Equal("A")), gomega.HaveField("GetName()", gomega.Equal("B")))

	g.Expect([]valueobject{{"A"}, {"B"}}).To(e)   // works
	g.Expect([]pointerobject{{"A"}, {"B"}}).To(e) // fails
}
```

thediveo:
looking into the matcher code base that was originally deliberate; if there is no reason why we should keep this restriction then there's a PR. Yeah, I've just run into this myself 😀

PR #802, opened 2024-12-06
resolves onsi/gomega#696: make HaveField great on pointer receivers…

thediveo:
…given only a non-addressable value
- don't reject pointer receivers on non-addressable values anymore; instead, create a pointer value to the original value and then lookup the pointer receiver method.
- switches corresponding unit test from rejection to acceptance.


onsi:
sweet!  and thanks for tackling the various open issues! ❤️❤️❤️

Issue #803, opened 2024-12-09
gmeasure.Sample does not wait for running jobs when run in parallel


MaximeWeyl:
gmeasure.Sample does not wait for running jobs when run in parallel. It launches the last jobs and then exits as soon as the conditions are met, without waiting for the last running jobs to finish.

Here is a simple test suite to showcase the issue.

The first test would be expected to pass, but does not.
The second test is a workaround that make it pass.

```go
package testginkgosample_test

import (
	. "github.com/onsi/gomega/gmeasure"
	"sync"
	"testing"
	"time"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

func TestTestginkgosample(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Testginkgosample Suite")
}

var _ = Describe("gmeasure.Sample", func() {
	var N int

	BeforeEach(func() {
		N = 12
	})

	It("should run all started functions", func() { //but fails
		exp := NewExperiment("Testginkgosample")

		var counter int
		var counterLock sync.Mutex

		exp.Sample(func(i int) {
			time.Sleep(1 * time.Second)
			counterLock.Lock()
			counter++
			counterLock.Unlock()
		}, SamplingConfig{
			N:           N,
			NumParallel: 5,
		})

		Expect(counter).To(Equal(N))
	})

	It("actually needs a wait group", func() {
		exp := NewExperiment("Testginkgosample")

		var counter int
		var counterLock sync.Mutex
		var wg sync.WaitGroup

		exp.Sample(func(i int) {
			wg.Add(1)
			time.Sleep(1 * time.Second)
			counterLock.Lock()
			counter++
			counterLock.Unlock()
			wg.Done()
		}, SamplingConfig{
			N:           N,
			NumParallel: 5,
		})

		wg.Wait()

		Expect(counter).To(Equal(N))
	})
})

```

PR #804, opened 2024-12-09
gmeasure.Sample does not wait for running jobs when run in parallel

MaximeWeyl:
Fix https://github.com/onsi/gomega/issues/803


onsi:
welp!  that's a pretty big oversight.  thanks for fixing it!

PR #805, opened 2024-12-15
Fix "closing server mulitple times" test in ghttp

stephenwithav:


thediveo:
The original test is correct, please see also https://onsi.github.io/gomega/#panic. In fact, your change won't even test, as the test cannot be build: you call `Close()` which returns no value(s) and thus can never pass this to `Expect` that needs one or more values.

I'm closing this PR as invalid. Please always run Gomega's own checks on itself before a PR (as I can tell from my own experience).

Issue #818, opened 2025-01-15
go.mod contains an explicit toolchain 


cwedgwood:
`go.mod` contains and explicit `toolchain`.

This is not needed and probably added by accident when `go.mod` requirements for newer Go were updated.

This slows down builds and in some cases will result in build failures (because the exact version might not be available).

PR #819, opened 2025-01-15
remove spurious "toolchain" from go.mod

cwedgwood:
Having go.mod specific a exact toolchain is not required and breaks some build systems.  The preference is to remove this allow Go to use the appropriate installed tooling.

Resolves: #818

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/onsi/gomega
 
 go 1.22.0
 
-toolchain go1.22.10
-
 require (
 	github.com/google/go-cmp v0.6.0
 	github.com/onsi/ginkgo/v2 v2.22.2


hem2 (approved):

cwedgwood:
> With https://go.dev/doc/toolchain as the authoritative documentation on toolchain semantics, all interpretation errors are thus to blame on me.
> 
> The toolchain mechanism has been introduced as of Go 1.21(.0). Gomega follows the N, N-1 support policy which at this moment resolves to 1.23.x and 1.22.x. It's debatable whether the toolchain should be 1.22.10, 1.22.0, or maybe 1.21.0 to better allow for some skew. But to my limited understanding we cannot keep a toolchain definition from Gomega's go.mod anymore, more so as our dependencies also now introduce toolchain definitions of their own.
> 
> You can always enforce your local toolchain with `GOTOOLCHAIN=local`, but if it is outside the lower limit of the support range this will fail. Still, you have some control, in particular, the choice of 1.22.x or 1.23.x if I understand the situation correctly.

Thanks, I'm aware of the background to `toolchain` and how to avoid some issues; but nonetheless it's not needed here and only showed up when it caused a regression in an internal system.

I think the correct and safe thing to do it remove it.

Thanks!

blgm (approved):
I think we are moving into a world where `toolchain` will be standard in `go.mod` files, but as it's not adding a lot of value here, and it's breaking someone, I'm happy to remove it for now. Though I would expect it to get added back at some point in the future, as the Go tooling sometimes adds it, and it would not be wrong to merge a change that added it.

Issue #837, opened 2025-04-01
Eventually / gomega.AsyncAssertion "has no field or method To"


grosser:
I'm using `.To(BeTrue())` etc for all my other assertions, just Eventually does not accept it, can that b added as an alias ?

```
Eventually(func() bool {
  return true
}, "1s", "10ms").To(BeTrue())
```

I know it reads a bit weird, but having the 1-off .Should stick out is also weird :D

onsi:
hey @grosser - I don't feel strongly either way.  Would you be up for submitting a PR?  No need to update the docs  - just a comment on the `AsyncAssertion` interface that the these are equivalent should suffice.  For completeness you'll want to add `To`, `ToNot` and `NotTo`.

PR #838, opened 2025-04-02
add To/ToNot/NotTo aliases for AsyncAssertion

grosser:
fixes fixes https://github.com/onsi/gomega/issues/837

```
go mod edit -replace=github.com/onsi/gomega=github.com/grosser/gomega@grosser/to
```

```
Eventually(func() bool {
	err := getObj(hpa)
	return err == nil && hpa.Spec.MaxReplicas == expectedReplicas
}, "1s", "10ms").To(BeTrue())
```

even better would be `ExpectEventually ... To BeTrue`, but 🤷

### DIFF
--- a/internal/async_assertion.go
+++ b/internal/async_assertion.go
@@ -145,10 +145,22 @@ func (assertion *AsyncAssertion) Should(matcher types.GomegaMatcher, optionalDes
 	return assertion.match(matcher, true, optionalDescription...)
 }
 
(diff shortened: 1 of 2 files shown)


onsi:
> even better would be `ExpectEventually ... To BeTrue`, but 🤷 

or even `Expect(foo).ToEventuallyBe(True())` but... 🤷 indeed.

onsi:
i've pulled this in and will cut a release so you can play with it.

onsi:
out now in 1.37.0

PR #848, opened 2025-06-18
gstruct: handle extra unexported fields

MaximeWeyl:
solves https://github.com/onsi/gomega/issues/849

reflect is not able to extract unexported fields, which makes gstruct unable to work with unexported struct fields.
Sometimes, we may want to be able to ignore them, because it makes sense to test only the exported fields (testing the exported API of a package, not the internal details). It works with the IgnoreExtras option, but this is IMO too powerful, because now you cannot detect ExportedFields that are later added to the struct but not to the test. 

This PR is a proposition to allow unexported fields of a struct to be ignored via : 
- using gstruct.Ignore() (since this matcher does not need an actual value, it can be called without retrieving the real actual value)
- using the new IgnoreUnexportedExtras option, that will ignore extras only if they are unexported

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2786,7 +2786,14 @@ Due to the global nature of these methods, keep in mind that signaling processes
 
 ### Testing type `struct`
 
-`gstruct` provides the `FieldsMatcher` through the `MatchAllFields` and `MatchFields` functions for applying a separate matcher to each field of a struct:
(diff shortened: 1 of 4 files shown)


onsi:
looks great!  thanks for the PR

MaximeWeyl:
My pleasure

Issue #876, opened 2025-10-08
MatchError cannot be used as is to ensure a particular sentinel error is returned.


Smerom:
Due to MatchError performing a deep equality check after testing for `errors.Is`, certain mistakes can slip past it when a test is wanting to ensure that a particular sentinel error is returned. As a simple example

```
var SentinelError = errors.New("some error")

Expect(errors.New("some error")).To(MatchError(SentinelError)) // passes

errors.Is(errors.New("some error"), SentinelError) // returns false
```

Which means if a function should, but fails to return a sentinel error, but returns an error with the same underlying string value, MatchError will not catch the bug.

Related to #452 and #439

thediveo:
can this issue be closed, based on the discussion in the above referenced abandoned PR #922?

PR #877, opened 2025-10-08
Add matcher reflecting errors.Is behavior

Smerom:
The current `MatchError` matcher considers errors to be a match that `errors.Is` does not, and so cannot be used in a test to ensure that upstream error handling can use `errors.Is` to for error specific handling. In order to not break any existing tests, I propose adding the following stricter error matcher that considers two errors a match if and only if they are both non-nil, and `errors.Is(actual, expected)` returns true.

As an example of current behavior:

`Expect(errors.New("some error")).To(MatchError(errors.New("some error")))`

will pass, however:

`errors.Is(errors.New("some error"), errors.New("some error"))` 

will return false.

Related to issue: #876

### DIFF
--- a/matchers.go
+++ b/matchers.go
@@ -146,6 +146,24 @@ func MatchError(expected any, functionErrorDescription ...any) types.GomegaMatch
 	}
 }
 
+// MatchErrorStrictly succeeds iff actual is a non-nil error that matches the passed in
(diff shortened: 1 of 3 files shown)


onsi (commented):
LGTM, thanks!

Issue #584, opened 2022-09-01
[nice to have] Would be very nice to have a `BeAnArray` matcher


eriknyk:
Would be very nice to have a `BeAnArray` matcher since we have similar Matchers like

.Should(BeADirectory())
.Should(BeARegularFile())
.Should(BeEmpty())

`Would be Nice`

```
.Should(BeAnArray(...))
```

This can be helpful when a variable has type `any` = `interface{}` it can be help to identify if the variable type is an expected array

---

Best Regards.

onsi:
Hey there - happy to accept a PR for this if you have the bandwidth to put one together.

sanbricio:
Hello!  @onsi  I've created a MR adding these interesting matchers! happy to colaborate to these project.

PR #906, opened 2026-04-12
Add BeASlice and BeAnArray matchers

sanbricio:
## Summary

Implements `BeASlice()` and `BeAnArray()` matchers as requested in #584.

These matchers enable type assertions for slice and array types when working with `interface{}` values:

- **BeASlice()**: Succeeds if actual is a slice type (`[]T`)
- **BeAnArray()**: Succeeds if actual is an array type (`[N]T`)

## Examples

```go
var data any = []int{1, 2, 3}
Expect(data).Should(BeASlice())

var arr any = [3]int{1, 2, 3}
Expect(arr).Should(BeAnArray())

Expect("hello").ShouldNot(BeASlice())  // string is not a slice

```

## Implementation

- Added BeASliceMatcher and BeAnArrayMatcher structs in matchers package
- Added constructor functions BeASlice() and BeAnArray() in root matchers.go
- Full test coverage including:
  - Multiple slice/array element types
  - Nil slice handling
  - Type validation (error on nil literal)
  - Failure message verification
  - Negation tests

## Testing
All 524 matchers tests pass
go vet ./... passes with no warnings
9 new specs covering both matchers

### DIFF
--- a/matchers.go
+++ b/matchers.go
@@ -621,6 +621,18 @@ func BeADirectory() types.GomegaMatcher {
 	return &matchers.BeADirectoryMatcher{}
 }
 
+// BeASlice succeeds if actual is a value of slice type.
+// This is useful when actual has type any (interface{}) and you want to assert it is a slice.
+func BeASlice() types.GomegaMatcher {
+	return &matchers.BeASliceMatcher{}
+}
+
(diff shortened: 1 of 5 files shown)


onsi (approved):
sorry for the delay - thanks for this!

PR #912, opened 2026-05-12
fix(format): detect pointer cycles to avoid runaway formatting output

### LINKED ISSUE #830: Formatting Cyclic Types produces infinite output
When using gomega (and, by extension, ginkgo) with cyclic types, the output from formatting the objects from e.g. `Expect(foo).To(Equal(bar))` is infinite, and if the cyclic pointer field is early in the struct, the actual differing fields may never actually be printed. This renders this output nearly useless.

I would like to propose and implement adding a new struct tag `gomega:"omit"` which replaces the formatting output of that field with a message indicating it was omitted, plus the address if it was a pointer. This would allow cyclic types to be formatted finitely.

SAY-5:
Per the discussion in #830, format.Object now tracks pointer addresses it has descended into during a single invocation and prints `0x... (cyclic reference)` instead of recursing again, so cyclic data structures format finitely and the differing fields stay visible. Closes #830.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -262,7 +262,7 @@ func Object(object any, indentation uint) string {
 	if err, ok := object.(error); ok && !isNilValue(value) { // isNilValue check needed here to avoid nil deref due to boxed nil
 		commonRepresentation += "\n" + IndentString(err.Error(), indentation) + "\n" + indent
 	}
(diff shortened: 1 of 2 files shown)


onsi:
fantastic thank you!

PR #922, opened 2026-07-23
fix(MatchError): match error values only via errors.Is

### LINKED ISSUE #876: MatchError cannot be used as is to ensure a particular sentinel error is returned.

Solaris-star:
## Summary

`MatchError(err)` previously fell back to `reflect.DeepEqual` along the unwrap chain. Two independently constructed `errors.New(\"same text\")` values are DeepEqual but not the same sentinel, so tests could pass even when production code failed `errors.Is` (#876).

## Fix

When the expected argument is an `error`, match only with `errors.Is`. String and matcher forms are unchanged.

## Tests

Updated MatchError matcher specs for sentinel-only semantics.

```text
go test ./matchers -count=1
# ok
```

Fixes #876

### DIFF
--- a/matchers/match_error_matcher.go
+++ b/matchers/match_error_matcher.go
@@ -29,17 +29,10 @@ func (matcher *MatchErrorMatcher) Match(actual any) (success bool, err error) {
 	expected := matcher.Expected
 
 	if isError(expected) {
-		// first try the built-in errors.Is
-		if errors.Is(actualErr, expected.(error)) {
-			return true, nil
-		}
-		// if not, try DeepEqual along the error chain
-		for unwrapped := actualErr; unwrapped != nil; unwrapped = errors.Unwrap(unwrapped) {
-			if reflect.DeepEqual(unwrapped, expected) {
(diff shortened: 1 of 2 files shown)


onsi:
hey sorry for the footgun.  you should use `MatchErrorStrictly` - unfortunately the PR would break backwards compatibility so I can't pull it in.

Solaris-star:
Thanks for the quick feedback! Understood — closing in favor of `MatchErrorStrictly` for sentinel-only matching. Appreciate the pointer.

Issue #451, opened 2021-06-16
any thoughts on adapting gomega matchers to gomock.Matcher interface


cadina:
so we can use them as gomock EXPECT arguments. the original matchers of gomock is too weak when dealing with complex types and assertions.

cadina:
now i'm using a simple wrapper like:

```
type GomegaMatcher interface {
	Match(actual interface{}) (success bool, err error)
	FailureMessage(actual interface{}) (message string)
}

type matcher struct {
	GomegaMatcher
	x interface{}
}

func (m matcher) Matches(x interface{}) bool {
	m.x = x
	result, _ := m.Match(x)
	return result
}

func (m matcher) String() string {
	return m.FailureMessage(m.x)
}

func m(gmather GomegaMatcher) gomock.Matcher {
	return matcher{gmather, nil}
}
```

onsi:
I was going to ask if a wrapper was working.  That would be what I would recommend and how I would approach it.

I could see adding a gomock adapter extension to gomega (something like `gomega/extensions/gomockadaptor`) - but i'd prefer not making this part of the core DSL or gomega matcher interface.

cadina:
> I was going to ask if a wrapper was working. That would be what I would recommend and how I would approach it.
> 
> I could see adding a gomock adapter extension to gomega (something like `gomega/extensions/gomockadaptor`) - but i'd prefer not making this part of the core DSL or gomega matcher interface.

that would be perfect.

i'm also thinking about taking full advantage of mockgen or something like that. just like how it generates mock recorders, we should be able to extract struct types from input and output arguments and generate strong typed structure matchers. that will a lot better than the map we used for `MatchAllFields`.

onsi:
i'm heads down working on ginkgo 2.0 these days but making it easier to build custom marchers in gomega is something i'm interested in working on next.  if you have bandwidth and are up for a PR I'd be happy to to pull in a gomock adapter and to take a look at a matcher generator with you!

cadina:
@onsi sure. i will take some time to walk through the codebase and see what i can do.

mgnsk:
Sorry to bump an old issue but I'm using this kind of an adapter:
```go
type gomegaAdapter struct {
	matcher        types.GomegaMatcher
	failureMessage string
}

func Matches(matcher types.GomegaMatcher) gomock.Matcher {
	return &gomegaAdapter{
		matcher: matcher,
	}
}

[...]

PR #923, opened 2026-08-25
Add gomock adaptor extension for using Gomega matchers with gomock

ChrisJr404:
Picks up #451. @onsi, on that issue you mentioned you'd be happy to pull in a gomock adapter as an extension package, so here it is.

It adds `extensions/gomockadaptor`. `Match` wraps a Gomega matcher in a value that satisfies gomock's `Matcher` interface (`Matches(any) bool` plus `String() string`), so any Gomega matcher can be used as a gomock argument matcher:

```go
mock.EXPECT().Send(gomockadaptor.Match(HaveLen(3)))
```

On a failed match it captures the matcher's `FailureMessage` and returns it from `String()`, so gomock's report explains why the argument didn't match. The package doesn't import gomock (it relies on structural typing), so it adds no dependencies to Gomega. The tests assert the adaptor satisfies the gomock Matcher shape via a local interface, so they stay dependency free too.

I kept it out of the core DSL like you suggested. Happy to rename or move the package if you'd prefer a different home for it.

### DIFF
--- a/extensions/gomockadaptor/gomockadaptor.go
+++ b/extensions/gomockadaptor/gomockadaptor.go
@@ -1,0 +1,64 @@
+/*
+package gomockadaptor lets you use Gomega matchers as gomock argument matchers.
+
+gomock (both github.com/golang/mock/gomock and go.uber.org/mock/gomock) accepts
+any value implementing its Matcher interface:
+
+	type Matcher interface {
+		Matches(x any) bool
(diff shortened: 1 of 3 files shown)


onsi:
thanks!  lgtm!

onsi:
just shipped in 1.43.0 - thanks again!

PR #924, opened 2026-09-17
dependencies: bump go.yaml.in/yaml/v3 v3.0.4 -> v3.0.5

pohly:
The main advantage is that the new version has fewer dependencies, which directly reduces the indirect dependencies of gomega. The old dependency tree of the v1.43.0 release tag was:

    $ go mod why github.com/rogpeppe/go-internal/fmtsort
    # github.com/rogpeppe/go-internal/fmtsort
    github.com/onsi/gomega/matchers
    go.yaml.in/yaml/v3
    go.yaml.in/yaml/v3.test   <=== pohly: these are the yaml unit tests
    gopkg.in/check.v1
    github.com/kr/pretty
    github.com/rogpeppe/go-internal/fmtsort

Now that dependency will be gone. The master branch still depends on it through ginkgo.


onsi:
shipped!

